PR #2, opened 2020-06-04
Update .gitignore

mbifulco:
Update the project's .gitignore file to include the defaults from gatsby, as well as the community created defaults for VS Code, MacOS, Windows, and Node projects from [gitignore.io](https://gitignore.io)


mbifulco:
One more quick commit here - I added yarn.lock to the gitignore, since it looks like we're using npm's `package-lock.json`

.gitignore
```diff
@@ -1,3 +1,224 @@
...
+
+# Output of 'npm pack'
+*.tgz
+
+# Yarn Integrity file
+.yarn-integrity
+
+# dotenv environment variables file
+.env
+.env.test
```
magnificode, line 165: Can we add a couple more dotenv patterns here:

`.env.development`
`.env.production`

These are specific patterns used by Gatsby for Environment Variables.

The rest looks super comprehensive, thank you!

mbifulco: @magnificode that'll be handled by `.env.*` on line 56 - sorry they're not right next to each other.  Lots of this boilerplate came from [gitignore.io](https://gitignore.io)

mbifulco: ...actually in thinking about this, that may not be a good thing to include in the gitignore.  We'll probably want a `.env.example` as part of the repo to show _all_ of the ENVs we're using.  I'm going to push that change shortly

magnificode:
@mbifulco 💯 - thank you!

PR #7, opened 2020-06-04
feat: setup airtable integration

jbolda:
# Describe your PR
It isn't pretty, but it is functional. It uses dotenv and requires an Airtable API key with access to the base. Each table in the base has it's own query through the `separateNodeType` option in the config and specified via the `queryName`.

To note, we have added Approved views in each of the tables. Any record that is approved will be pulled in through our queries and placed on the website.


magnificode:
Note to self :: Make sure the Airtable API is added to Netlify env :)

PR #8, opened 2020-06-04
feat: Add cloudinary hooks + serverless function

domitriusclark:
# Describe your PR
To handle our media I've installed [use-cloudinary](https://use-cloudinary.netlify.app/) into our project for client + serverless media management via Cloudinary

You'll now see:

- A new `Image` component. This will be how we'll call Image assets from Cloudinary. You can also add any Chakra Image component styles 👍
- A new `functions` directory, which will house all of our serverless functions via Netlify. So far we'll be only using a serverless function for hitting a signed search via the `Cloudinary` Node SDK
- A new `Search` component. This will serve as an early example of how to utilize the useSearch hook w/ the above endpoint to return a group of media
- changes to `netlify.toml` to include new serverless function commands to the build step 
- new doc area for Cloudinary specific env setting


functions/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "functions",
+  "version": "1.0.0",
+  "description": "",
+  "main": "upload.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
```
mbifulco, line 7: - [ ] We should create a card to set up tests for these serverless functions

mbifulco:
@domitriusclark I'm seeing an eslint warning for the `useEffect you've got on `Image.js` - 

image

Is this effect meant to be run only once? If that's the case, can you add 

```es6
/* eslint-disable react-hooks/exhaustive-deps */

React.useEffect( /* etc */)

/* eslint-enable react-hooks/exhaustive-deps */
```
around the effect?

src/components/Image.js
```diff
@@ -0,0 +1,29 @@
...
+    getImage({
+      public_id: publicId,
+      transform_options: {
+        ...transforms
+      }
+    })
+  }, [])
+
+
+  // Can make use of the `Spinner` or `Skeleton` component from Chakra UI here
```
magnificode, line 22: Mind turning these into @TODO comments? I'll comb the codebase periodically for them and make sure they're added to Trello.

domitriusclark: Most of my comments here were specifically for review purposes since the lib is fairly new. 

I'd be happy to do that though and make them actionable. Could you show me an example of what you're looking for?

domitriusclark: Oh that's rad. What the extension 👀

magnificode: https://marketplace.visualstudio.com/items?itemName=wayou.vscode-todo-highlight

It's got a nice little setting where you can highlight any keyword you'd like 😄

PR #13, opened 2020-06-05
Add CI test runner for PRs

ChrisBrownie55:
Added in a GitHub Action that will run all tests via the `npm test` command on any PRs. If you would like it to only run tests on PRs that are merged with `next` and `master` branches. You could do:

```yaml
on:
  pull_request:
    branches: [ master, next ]
```

A short synopsis of what this does:

```bash
# Someone made a PR or pushed a change to a PR
# GitHub starts up this Action:

# installs dependencies
npm install

# runs tests
npm test
```

Should we add `npm test:e2e` to this as well?


ChrisBrownie55:
We could also easily add in an `npm lint`

mbifulco:
This is awesome! I feel like lint and prettier would be a good idea to add to this, too.  It's possible to extend eslint settings to run prettier tests - I can see about setting that up in a separate PR

ChrisBrownie55:
@mbifulco I don't think adding in running prettier during this would be a good idea since we already have a precommit hook that runs prettier on staged files anyway

PR #15, opened 2020-06-05
a11y improvements for site navigation

chancestrickland:
This PR makes a few minor but important adjustments for better accessibility.
- Fixed some invalid HTML in the list (missing `li` items)
- Added a proper button to the nav toggle with a visually hidden label
- Added a skip nav link for better keyboard/screen reader navigation

It also makes some small tweaks to the header so that the menu is toggled correctly when switching the screen between small and medium widths. Lastly, I broke out `PrimaryNav` from `Nav` so that `Nav` exports more generally useful nav menu components we can use elsewhere on the site.

Changes are not visual for the most part, so I'm not sure that a video or screens are helpful. The main exception being that when you tab from the top of the document, you will land on a "Skip to content" link that is otherwise hidden.


src/components/Nav.js
```diff
@@ -0,0 +1,57 @@
+import React, { forwardRef } from 'react';
+import { Flex, Link, Box, List, ListItem } from '@chakra-ui/core';
+import { Link as GatsbyLink } from 'gatsby';
+
+const Nav = forwardRef(
```
domitriusclark, line 5: TIL more about `forwardRef` 👍

src/components/Nav.js
```diff
@@ -0,0 +1,57 @@
+import React, { forwardRef } from 'react';
+import { Flex, Link, Box, List, ListItem } from '@chakra-ui/core';
```
domitriusclark, line 2: `Box` is unused here -- think this was just missed from initial

chancestrickland: @domitriusclark Removed

magnificode:
Excellent work @chancestrickland - thank you!

PR #18, opened 2020-06-05
Add a "lint" command to package.json for CI

mbifulco:
# Describe your PR
This PR sets up and configures eslint for this project, per guidelines from [gatsby](https://www.gatsbyjs.org/docs/eslint/), [prettier](https://prettier.io/docs/en/integrating-with-linters.html), and [cypress](https://github.com/cypress-io/eslint-plugin-cypress).

Related to #13 
Fixes #

Just CI stuff

## Steps to test

1. run `npm run lint` to see the lint test in action

### TODO
- [x] (**all set - thanks @jakeleboeuf!**) I'm not a Cypress expert, and could use some help determining what's going on with eslint errors coming from existing cypress tests around this rule:

`Do not wait for arbitrary time periods  cypress/no-unnecessary-waiting`

At present, if you run `npm run lint` on this branch, you'll see a handful of those. [Cypress docs](https://docs.cypress.io/guides/references/best-practices.html#Unnecessary-Waiting) seem to indicate that our test code doesn't need to include wait commands, but I've never used it before - I'd like a gut check on that 👍


mbifulco:
Sorry, my TODO description above was wrong - this PR currently has a rule set up to ignore the cypress warning, so those errors won't show when lint is run.  That can be changed by commenting out the rule on line 12 of `eslintrc.js`

.eslintrc.js
```diff
@@ -0,0 +1,14 @@
...
+    __PATH_PREFIX__: true,
+  },
+  extends: [
+    'react-app',
+    'plugin:cypress/recommended',
+    'plugin:chai-friendly/recommended',
+    'plugin:prettier/recommended',
+  ],
+  rules: {
+    'cypress/no-unnecessary-waiting': [0],
```
jakeleboeuf, line 12: @mbifulco I've had plenty of times where defining a `wait()` of some kind was useful to simulate some user behavior- but that is a rare case. I'd say leave that rule on as a `warning` for now and we can change it if it becomes an issue.
```suggestion
    'cypress/no-unnecessary-waiting': 'warn',
```

mbifulco: Happy to defer to your wisdom here.  I'm good with that!

mbifulco:
Blah - looks like something went awry with Netlify.  I think an owner is going to need to clear the cache and rebuild to get this to pass, or I can push a meaningless commit

mbifulco:
Todo:
- [ ] the following should be added to `.github/workflows/node.js.yml` to run lint during CI:

```yml
      - name: Run ESLint
        run: npm run lint
```

racedale:
> edit: for clarity, I can't push a commit with that addition because I don't have `workflow` scope on this Repo
> 
> ```
>  ! [remote rejected] feature/lint-prettier -> feature/lint-prettier (refusing to allow an OAuth App to create or update workflow `.github/workflows/node.js.yml` without `workflow` scope)
> ```

Huh, okay I can push the change for you

ChrisBrownie55:
Oh also we should probably change the file name to be more apt, that's my fault

ChrisBrownie55:
I think instead of `node.js.yml` it should probably be something like `pull-request.yml`, `pr-checker.yml`, or `test-pr.yml`

racedale:
I'm going to merge this, we can always rename the file in another PR

PR #19, opened 2020-06-05
feat: adds lang attr to Helmet component.

chrisdemars:
# Describe your PR

This change adds an `en` lang attribute to the `Helmet` component in the `index.js` page. Currently we have zero violations now after running aXe. It was the only violation I saw at the moment.

### DIFF
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -9,6 +9,7 @@ export default () => {
     <Layout>
       <Helmet>
         <title>Home</title>
+        <html lang="en" />
       </Helmet>
       <Flex align="center" justify="center">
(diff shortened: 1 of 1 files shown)


src/pages/index.js
```diff
@@ -9,6 +9,7 @@ export default () => {
     <Layout>
       <Helmet>
```
magnificode, line 10: We should be able to leverage the `htmlAttributes` attribute on the `<Helmet>` tag.

`htmlAttributes={{ lang: 'en' }}` should do the trick!

chrisdemars:
Reference this PR for code changes.

https://github.com/Rebuild-Black-Business/RBB-Website/pull/36

PR #23, opened 2020-06-05
Feature: Button Components

andystitt829:
I built the button components requested in this <a href="https://trello.com/c/j6uYMlkL/29-button-components">Trello ticket</a> using the specs in this <a href="https://www.figma.com/file/Lo1ZcBOO9t4QRbX2I0PwBQ/RBB?node-id=2%3A102">Figma design</a>.

I used the PseudoBox component from Chakra UI since it easily allows to style for various button states. I added the text prop to it so that any of the buttons can be used anywhere like `<CTAButton text="Text here" />`.

I don't use React in my every day work (I use WordPress and Bootstrap at my government contract job) and have only used it to create my portfolio site in Gatsby. If there's a better way of doing any of this, feel free to let me know.


magnificode:
@andystitt829 Thank you so much for the PR here!

This isn't documented super well so it's on me, but we do have a custom Chakra theme that employs the use of variable names for the color pallet defined here: https://www.figma.com/file/Lo1ZcBOO9t4QRbX2I0PwBQ/RBB?node-id=2%3A6

Some of the colors in our Chakra theme may be out of sync with the above Figma link, but have since been finalized. Could you update the buttons here to leverage the colors from our Chakra theme located here: https://github.com/Rebuild-Black-Business/RBB-Website/blob/master/wrap-root-element.js

andystitt829:
@magnificode Thanks so much for the feedback. I'm unable to find documentation that shows me how to implement those variables and have them actually work. I'm putting the "rbb-orange" variable in quotes in place of the hex value, but it's not working. The Chakra docs aren't showing me what I need, so I'm guessing this is a JSX syntax thing that I'm not aware of? Would love some assistance.

ChrisBrownie55:
@andystitt829 I hadn't seen this and, by accident, had ended up making my own PR doing the same thing except using Chakra's useTheme hook for easy configuration and `<Button>` (for optional added functionality). And I was wondering if you wanted to take a look at that and compare the two. Again, my apologies I hadn't seen that you had claimed this on Trello.

PR #27, opened 2020-06-05
Chore - Update gitignore with WebStorm

ChrisBrownie55:
## Added WebStorm related files to .gitignore

I updated the `.gitignore` file to include WebStorm related files (such as `.idea/`). And I used the link to gitignore.io in the file to update it so it should be good to go 👍

If anyone needs to add anything to the gitignore feel free to commit.


mbifulco:
@ChrisBrownie55 I think your changes removed `.env.development` and `.env.production` - that's on me -- those aren't generated by gitignore.io, and I added them in the middle of the file.  Can you drop them in at the beginning or end of the file with a comment to explain that they're not generated?

ChrisBrownie55:
@mbifulco I fixed it (I also added back in `public/`) thanks for catching that 👍

racedale:
We should also add 
```
cypress.env.json
cypress/videos
cypress/screenshots
```

Issue #24, opened 2020-06-05
Create filtering and sorting logic


magnificode:
### Business Filtering

![Screen Shot 2020-06-05 at 11 08 43 AM](https://user-images.githubusercontent.com/1380536/83904346-ed470980-a71c-11ea-9605-7c5ba60f4129.png)

Filters should be created to allow a user to filter results from the [BusinessInNeedFeed](https://github.com/Rebuild-Black-Business/RBB-Website/blob/master/src/components/Feeds/BusinessInNeedFeed.js) component by the following criteria:

- Business Type
  - Label: "Business"
  - Type: single select
    - Options: "Entertainment", "Food and beverage", "Health and wellness", "Professional services", "Retail"
  - Placeholder: "Select type"

- Location
  - Label: "Location"
  - Type: text
    - Allow users to type their city. i.e "Denver" should return any business who's location includes "Denver".
  - Placeholder: "Enter city"

- Need
  - Label: "Need"
  - Type: single select
    - For V1 this should be a "disabled" field that pre-populates with an "Impacted by protests" option.
    - For V1 this select field _will not_ affect filtered results. The business filters will _only_ display businesses in need.

- Submit
  - Type: submit
  - Label: "Search"
  - On click of the submit button **or** on 'enter', the filtered results should appear below the filters without a page refresh.

### Resource Filtering

![Screen Shot 2020-06-05 at 11 19 21 AM](https://user-images.githubusercontent.com/1380536/83905147-698e1c80-a71e-11ea-95a7-03d8ca16a4fd.png)

Filters should be created to allow a user to filter results from the [SupportingOrgFeed](https://github.com/Rebuild-Black-Business/RBB-Website/blob/master/src/components/Feeds/SupportingOrgFeed.js) component by the following criteria:

- Resources For
  - Type: single select
    - Options: "Ally", "Business owner"
  - Label: "Resources for"
  - Placeholder: "Who are you?"

- Location
  - Label: "Location"
  - Type: text
    - Allow users to type their city. i.e "Denver" should return any business who's location includes "Denver".
  - Placeholder: "Enter city"

- Submit
  - Type: submit
  - Label: "Search"
  - On click of the submit button **or** on 'enter', the filtered results should appear below the filters without a page refresh.

### Ally Filtering
![71cb190824b43e77781da33212907e42](https://user-images.githubusercontent.com/1380536/83924050-d4e8e600-a740-11ea-830b-1d44277ad0c1.png)

[...]

PR #29, opened 2020-06-05
[feature] Add Resource filter

ryancharris:
# Describe your PR

This PR adds a new component called `<ResourceFilter />`, which is rendered on `/resources` and allows users to filter by:
1. Location
2. Category

The user will trigger a search and see new results whenever they:
1. Click on the `Search` button
2. Hit Enter while focused on the `Location` input
3. Hit Enter while focused on the `Search` button

The only page this changes right now is `/resources`, however, when adding the other filter types I may come back and abstract some of the filtering logic to a shared component. TBD on that 👍

https://www.loom.com/share/61483f4d42a84180b5a5f3f94bb3099a

## Steps to test

1. Enter a location in the input
2. Hit Enter to trigger a search
3. Alter the location input value
4. Press Tab to focus the Search button
5. Hit Enter to trigger a search
6. Alter the location input
7. Click the Search button to trigger a new search


ChrisBrownie55:
@ryancharris Added a comment on #24 with a function that should help with filtering. Tested locally so it should work just fine for you 👍

I would suggest using `useMemo()` with it since we'll likely end up having a lot of data to filter through and we don't want that to refilter that data on every re-render.

ryancharris:
> @ryancharris Added a comment on #24 with a function that should help with filtering. Tested locally so it should work just fine for you 👍
> 
> I would suggest using `useMemo()` with it since we'll likely end up having a lot of data to filter through and we don't want that to refilter that data on every re-render.

@ChrisBrownie55 Good call on using `useMemo`. I was just getting it to work with `useEffect` before making that improvement 👍

magnificode:
@ryancharris looking solid, thank you!

PR #33, opened 2020-06-05
Feature: ResultCard component

chancestrickland:
This PR creates a new `ResultCard` component as well as more generic composable `Card` components. Closes #22.

**IMPORTANT:** There are edits to the homepage that should not be merged. These are intended to make it quick to review these components locally.

Component API:

```js
<ResultCard
  // assumes these come from a list somewhere, most likely
  businessCategory="foodAndBeverage"
  businessName="Dad's Donuts"
  businessDescription="Blah blah blaaaah"
  businessLocation="Atlanta, GA"
  businessUrl="https://dobusiness.com"
  donationUrl="https://donatemoney.com"
  // overrides the default fallback image
  imageSrc="https://imagesrc.jpg"
  imageAlt="cool image you got there"
/>
```

### Screenshots / video of changes

image

## Steps to test

Run locally and check out the homepage.

PR #35, opened 2020-06-05
add basic schema markup to businesses

kkemple:
# Describe your PR
This adds the very beginning of schema markup to businesses so that they can be easily parsed by scrapers like Google.

There isn't a lot of data to go on so implemented what I can. One thing to call out is that the category in Airtable isn't going to match the category types from Schema.org. Once we know all the types we'll need a map. This is a good start though! 

Related to #
Fixes #

## Steps to test

1.

### DIFF
--- a/src/components/Feeds/BusinessInNeedFeed.js
+++ b/src/components/Feeds/BusinessInNeedFeed.js
@@ -3,28 +3,45 @@ import { StaticQuery, graphql } from 'gatsby';
 
 import { Box } from '@chakra-ui/core';
 
-const SupportingOrgs = data => {
-  return <Box as="pre">{JSON.stringify(data, null, 2)}</Box>;
+const Businesses = data => {
+  return (
+    <Box as="ul">
+      {data.allAirtableBizInNeed.edges.map(({ node: { data } }) => (
+        <Box as="li" itemScope itemType={`http://schema.org/${data.Category}`}>
+          {/* the category in Airtable isn't going to match the schema category, will need to be fixed */}
+          <h2 itemprop="name">{data.Name}</h2>
+          <h4 itemprop="address">{data.Location}</h4>
+          {/* it would be good to get more address details as well */}
+          <span itemprop="email">{data.Email}</span>
+          <a href={data.Donation_Website}>Donate</a>
+        </Box>
+      ))}
+    </Box>
+  );
 };
 
 export default props => (
   <StaticQuery
     query={graphql`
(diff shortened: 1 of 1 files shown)


magnificode:
@kkemple thank you for this! We'll use this as a jumping off point for the folks working on the various card types!

magnificode:
No idea why the deploy preview is failing however....looking into this.

src/components/Feeds/BusinessInNeedFeed.js
```diff
@@ -3,28 +3,45 @@ import { StaticQuery, graphql } from 'gatsby';
 
 import { Box } from '@chakra-ui/core';
 
-const SupportingOrgs = data => {
-  return <Box as="pre">{JSON.stringify(data, null, 2)}</Box>;
+const Businesses = data => {
+  return (
+    <Box as="ul">
+      {data.allAirtableBizInNeed.edges.map(({ node: { data } }) => (
```
magnificode, line 9: AH! I think this is failing on the frontend because the data returned from Airtable comes down with a `data` object.

`{data.data.allAirtableBizInNeed.edges.map....` seems to fix the issue @kkemple

kkemple: I'll update first thing in the morning! It was working locally so also confused!

magnificode: @kkemple I may be able to get the work you completed here wrapped in with the actual Card component scaffold here: https://github.com/Rebuild-Black-Business/RBB-Website/pull/33 if that's alright with you!

kkemple: Yeah totally! Didn't realize there was one!!

magnificode: @kkemple these folks are moving at an insane pace, so no worries at all. Just pinged that other ticket with your work here. Thank you again!

ChrisBrownie55: Technically speaking. It's a `data` prop that's being passed into the component. I think the real culprit here is this:

```jsx
// what was written:
const Businesses = data => {
  data.data.airtableStuff
}

// what that actually is:
const Businesses = props => {
  props.data.airtableStuff
}
```

magnificode:
Closing as this was rolled into #33 - Thanks again for laying the groundwork here @kkemple !

PR #46, opened 2020-06-06
Feature/zipcode utility function

brewswain:
# Zipcodes Utility Function
This PR creates a utility function that takes a zipcode as its input, and returns location information of said zipcode, inclusive of City and State in an object,

For this I used the 'zipcodes' library recommended, [which can be seen here](https://www.npmjs.com/package/zipcodes)

Simple test usage can be seen below:
image

Please note that the require statement was replaced with the ES6 standard import statement in the project, and the 'zipcodes' library was added as a devDependency.


brewswain:
Oops, let me fix these problems

package.json
```diff
@@ -63,7 +63,8 @@
     "lint-staged": "^10.2.8",
     "prettier": "^2.0.5",
     "react-test-renderer": "^16.13.1",
-    "start-server-and-test": "^1.11.0"
+    "start-server-and-test": "^1.11.0",
+    "zipcodes": "^8.0.0"
```
racedale, line 67: Since this is for the UI it will need to not be a devDependency

magnificode: Ah, blame me for that :)

brewswain: No problem, I'll get on that right now!

brewswain: Ok, done.

magnificode:
@brewswain looks good to me, thank you!!

src/utils/zipcodeConversion.js
```diff
@@ -0,0 +1,15 @@
+import zipcodes from 'zipcodes';
+
+const zipcodeConversion = zipcode => {
+  try {
+    let convertedZipcode = zipcodes.lookup(zipcode);
+    let error =
+      'Sorry, zipcode is invalid. Please check your input and try again.';
+    if (!convertedZipcode) throw error;
```
ChrisBrownie55, line 8: Should this not be:

```js
if (!convertedZipcode) throw new Error(error)
```

src/utils/zipcodeConversion.js
```diff
@@ -0,0 +1,15 @@
...
+    return convertedZipcode;
+  } catch (error) {
+    console.log(error);
```
ChrisBrownie55, line 11: Why are we just logging the error?
The error should probably be returned so it can be handled by the code using this utility function.

src/utils/zipcodeConversion.js
ChrisBrownie55, line 8: Should this not throw an actual error that also contains the zipcode that caused the error.
```js
if (!convertedZipcode) throw new Error(`Zipcode "${zipcode}" is invalid.`)
```

Issue #48, opened 2020-06-07
Create Pagination Component


racedale:
## Pagination
Figma: https://www.figma.com/file/Lo1ZcBOO9t4QRbX2I0PwBQ/RBB?node-id=579%3A1516

Leveraging Chakra UI components, we need to build out a Pagination component that can support having optional ellipsis for large amounts of pages of results. 

- Given a user is on the first page + 4 more (5 pages total), the first set of ellipsis will not be shown.
<img width="363" alt="Screen Shot 2020-06-06 at 10 47 19 PM" src="https://user-images.githubusercontent.com/6390640/83959954-95a5bc80-a848-11ea-8efe-badb834096d6.png">

- Given a user is on the last page - 4 less (5 pages total), the first set of ellipsis will not be shown. 
<img width="595" alt="Screen Shot 2020-06-06 at 10 54 10 PM" src="https://user-images.githubusercontent.com/6390640/83959958-aa825000-a848-11ea-8648-eb9e95bfa9f1.png">

- Given a user is somewhere in the "middle" of the results (not the previous two scenarios) then the 
middle number will be active and both sets of ellipsis will be shown
<img width="598" alt="Screen Shot 2020-06-06 at 10 46 31 PM" src="https://user-images.githubusercontent.com/6390640/83959871-9f7af000-a847-11ea-863b-5b027b06ac69.png">

Mobile:
<img width="613" alt="Screen Shot 2020-06-06 at 10 53 12 PM" src="https://user-images.githubusercontent.com/6390640/83959963-b66e1200-a848-11ea-808d-f10c2bc5ab54.png">

**If it's quick to add**, support desktop version as well. If not it's lower priority and should be created as a separate Issue, so notify @racedale & @magnificode 
Desktop:
<img width="688" alt="Screen Shot 2020-06-06 at 10 16 50 PM" src="https://user-images.githubusercontent.com/6390640/83959642-f6cb9100-a844-11ea-9336-3127628bd612.png">

racedale:
Trello card: https://trello.com/c/zxR8G4LA/44-create-pagination-component

billy-le:
I would like to take this one.

magnificode:
@racedale are we good to close this out since #49 is merged?

PR #49, opened 2020-06-07
feature/pagination-component

billy-le:
created pagination component and added pagination placeholders to business, allies, and resources pages

# Describe your PR
Requesting to merge Pagination Component to project

Related to #48 

Business, Allies, and Resources Pages

https://www.loom.com/share/899d1362e0d94006bb73b0dbcd2ab6a5


src/components/Pagination/Pagination.jsx
```diff
@@ -0,0 +1,208 @@
...
+      setPageLimit(pageLimit);
+    }
+    if (pageNeighbors) {
+      setPageNeighbors(Math.max(0, Math.min(pageNeighbors, 2)));
+    }
+
+    const totalPages = Math.ceil(totalRecords / pageLimit);
+    setTotalRecords(totalRecords);
+    setTotalPages(totalPages);
+  }, []);
```
ChrisBrownie55, line 42: Uhh, there's a warning here about not providing the `totalRecords`, `pageLimit`, and `pageNeighbors` values from `props` to the dependencies array. That should probably have been addressed before merging as that opens up the code to possible bugs in edge cases.

src/components/Pagination/Pagination.jsx
```diff
@@ -0,0 +1,208 @@
...
+
+  useEffect(() => {
+    setPages(fetchPageNumbers());
```
ChrisBrownie55, line 45: @billy-le This should definitely be addressed soon too. This is a major concern here.

`fetchPageNumbers()` should be declared within this `useEffect()` call so that you can properly declare it's dependencies which include `pageNeighbors`, `totalPages`, and `currentPage` among others.

src/components/Pagination/Pagination.jsx
```diff
@@ -0,0 +1,208 @@
...
+
+  const handleMoveRight = () => {
+    gotoPage(currentPage + pageNeighbors * 2 + 1);
+  };
+
+  return (
+    <Flex flexWrap="nowrap" justifyContent="center">
+      {pages.map((page, i) => {
+        return (
+          <PseudoBox
```
ChrisBrownie55, line 155: @billy-le @racedale Big a11y issue here:

This should definitely not be a `<PseudoBox>` since those render plain old `<div>`'s which is a huge accessibility concern. This should be a `<Button>` since it has an action attached to it (not a link since it doesn't change the URL).

src/components/Pagination/Pagination.jsx
billy-le: Thanks for your comments @ChrisBrownie55! I can work on a fix to address these errors but I believe someone is currently working on tweaking the pagination buttons.

I can wait until after they are done to see if they have already fixed it.

src/components/Pagination/Pagination.jsx
```diff
@@ -0,0 +1,208 @@
...
+      }
+
+      return [1, ...pages, totalPages];
+    }
+
+    return range(1, totalPages);
+  };
+
+  const gotoPage = page => {
+    const { onPageChanged = f => f } = props;
```
ChrisBrownie55, line 124: This should be in an if-statement below instead, no need to make a new function.

```
if (onPageChanged) onPageChanged(paginationData)
```

PR #53, opened 2020-06-07
update businessfeed to use ResultCard instead of raw json

M0nica:
# Describe your PR

"The design for the result cards has not been fully approved yet
We do however have a "ResultCard" component in master that can be hooked up
It's just a bare bones component, but will definitely look nicer than straight JSON "

Related to #
Fixes #

/businesses
/resources

Screen Shot 2020-06-07 at 2 56 10 PM
Screen Shot 2020-06-07 at 2 56 19 PM

## Steps to test

1. Go to /businesses and see the ResultCards being displayed
2. Go to /resources and see the ResultCards being displayed

### DIFF
--- a/src/components/Feeds/BusinessFeed.js
+++ b/src/components/Feeds/BusinessFeed.js
@@ -1,12 +1,21 @@
 import React from 'react';
 import { StaticQuery, graphql } from 'gatsby';
-
+import ResultCard from '../ResultCard';
 import { Box } from '@chakra-ui/core';
 
 const BusinessesFeed = data => {
   return (
     <Box as="pre" whiteSpace="break-spaces">
-      {JSON.stringify(data, null, 2)}
+      {data.data.allAirtableBusinesses.nodes.map(item => (
+        <ResultCard
+          name={item.data.Business_Name}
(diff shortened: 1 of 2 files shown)


src/components/Feeds/BusinessFeed.js
```diff
@@ -1,12 +1,21 @@
 import React from 'react';
 import { StaticQuery, graphql } from 'gatsby';
-
+import ResultCard from '../ResultCard';
 import { Box } from '@chakra-ui/core';
 
 const BusinessesFeed = data => {
+  const businesses = data.data.allAirtableBusinesses.nodes;
   return (
     <Box as="pre" whiteSpace="break-spaces">
```
magnificode, line 9: Since we're returning the ResultCard, instead of JSON, we can move this `<pre>` wrapper!

juanri0s:
LGTM

src/components/Feeds/BusinessFeed.js
M0nica: I removed the `<pre>` it seems like the actual `<Box>` is required to ensure that the width of all of the ResultCards is consistent.

PR #55, opened 2020-06-07
[feature] Add Business filtering

ryancharris:
# Describe your PR

This PR adds a new filtering component to `/businesses` that allows users to filter businesses by:
1. Type
2. Location (zip or city/state) string
3. Whether the business was effect by protests or not

The `BusinessFilter` component will only appear on the `/businesses` route for now.

https://www.loom.com/share/fc78caaba9a7474b86bc41ae71d6ce02

## Steps to test

1. Go to the `/businesses` route
2. Try out the filters
3. Confirm the data is being re-filtered


mbifulco:
Works like a charm @ryancharris - great job!

One suggestion, which may be good for discussion with the design team: I feel like it might make more sense to have the check box filter function a little differently, and _only_ show affected businesses when true (like it is now), but otherwise show _all_ businesses.

image

I can't imagine a valuable use case for someone wanting to see _just_ businesses who are unaffected by protests. 🤔 What do y'all think?

magnificode:
> Works like a charm @ryancharris - great job!
> 
> One suggestion, which may be good for discussion with the design team: I feel like it might make more sense to have the check box filter function a little differently, and _only_ show affected businesses when true (like it is now), but otherwise show _all_ businesses.
> 
> image
> 
> I can't imagine a valuable use case for someone wanting to see _just_ businesses who are unaffected by protests. 🤔 What do y'all think?

I'll +1 this, @ryancharris does this cause any big headache on your end?

Beyond that, this functions incredibly...thank you so much for busting this out!!

magnificode:
Worth noting as well #46 was merged in, it may be worth moving that uitlity function into your more generalized locationUtils.js file!

magnificode:
@ryancharris - Looks awesome, very much appreciate the hard work! And thanks for being flexible as requirements changed a bit!

Issue #50, opened 2020-06-07
Create Link Styles


racedale:
## Link Styles
Figma: https://www.figma.com/file/Lo1ZcBOO9t4QRbX2I0PwBQ/RBB?node-id=585%3A351

Leveraging Chakra UI components, we need to setup link styles for default, hover, active, an visited states. 

Note: The orange link should only be used for flagging businesses/allies

Mobile:
<img width="1075" alt="Screen Shot 2020-06-07 at 9 08 40 AM" src="https://user-images.githubusercontent.com/6390640/83971011-84879a80-a89e-11ea-94c7-efa082efadc2.png">

PR #56, opened 2020-06-07
Feature/link styles

subject026:
Link component combining gatsby and chakra link components with default and hover state styles.

Optional isCallToAction prop for applying specific call to action styles.

Related to #50 

default state:
links

standard link hover - can't see my mouse for some reason:
links2

call to action link hover again no mouse visible:
links3


subject026:
Screenshot from 2020-06-07 22-06-49

Looks like the footer link styles have been decided now and I can add another optional prop for this.

src/components/Link.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import { Link as ChakraLink, useTheme } from '@chakra-ui/core';
+import { Link as GatsbyLink } from 'gatsby';
+
+function Link(props) {
+  const theme = useTheme();
+  const isCallToAction = props.isCallToAction ? true : false;
```
mbifulco, line 7: I believe the preferred way to do this kind of check is
```suggestion
  const isCallToAction = (!!props.isCallToAction);
```

This has to do with _truthy_ and _falsy_ values in javascript, and will ensure that only boolean values work here.  As it's written right now, this would render as a CTA link, which is counterintuitive:

```jsx
<Link isCallToAction={5} />
```

subject026: Thanks that's a very good point! 

My thinking was that the component would be used like this so I was just checking whether the prop was present:

```jsx
<Link isCallToAction />
```
Maybe not good practice.

I'm now looking at the styles for links in the footer, maybe makes more sense to have a type prop and do something like this:

```jsx
<Link /> // no prop for default link
<Link type="callToAction" />
<Link type="footer" />
```
Do you think this would make more sense? Rather than

```jsx
<Link isCallToAction={true} />
<Link isFooterLink={true} />
```

racedale: Whenever I've written `<comp prop={true}>` in my projects it always gets auto corrected to `<comp prop>` due to [this linting rule](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-boolean-value.md)

subject026: > Whenever I've written `<comp prop={true}>` in my projects it always gets auto corrected to `<comp prop>` due to [this linting rule](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-boolean-value.md)

I think I've had this before myself actually...

It's late where I am right now so I'm going to bed. I'll work it out in the morning :)

ChrisBrownie55: @racedale and @subject026,
In the `<Button>`  component we are currently using a `variant` prop that takes in one of the three:
- cta
- primary
- secondary

So I think for this `variant` would be the correct way to go with the following options:
- cta
- footer

And the styles used to theme these links should probably be contained in the `src\gatsby-plugin-chakra-ui\theme.js` file.

src/components/Link.js
```diff
@@ -0,0 +1,36 @@
...
+
+  return (
+    <ChakraLink
+      as={GatsbyLink}
+      {...props}
```
ChrisBrownie55, line 12: This needs to go at the end of the component declaration to allow one to override the settings defined below if they need to.

Reasoning:

```jsx
function Example({ variant, ...props }) {
  return (
    <ChakraLink
      {...props} // <- this gets overridden by the next lines
      color={theme.links[variant].color.default}
      {/* Other styles */}
      {...props} // <- this allows the developer using this component to override if they need to
    />
  )
}
```

src/components/Link.js
subject026: @ChrisBrownie55 @mbifulco @racedale Thanks for the comments! 

Made some changes and logic is now more consistent with the Button component. 3 variants now to cover standard, cta and footer links.

subject026:
@ChrisBrownie55 @mbifulco @racedale Thanks for the comments!

Made some changes and logic is now more consistent with the Button component. 3 variants now to cover standard, cta and footer links.

magnificode:
@subject026 - this looks solid to me! Really appreciate the hard work, and good conversations here. Once @ChrisBrownie55 and @mbifulco give the thumbs up, I'll get this merged!

Issue #57, opened 2020-06-07
Ensure small screen Navigation is accessible


magnificode:
In our PrimaryNav component: https://github.com/Rebuild-Black-Business/RBB-Website/blob/master/src/components/PrimaryNav.js

Primarily I believe we need to add a couple of `aria` roles to the navigation (nav is only toggled on smaller screens).

`aria-controls` should be added to the `<button>` within the `<Box>` component wrapping the Hamburger menu. This should point to the ID of the navigation.

`aria-expanded` should be added to the `<button>` within the `<Box>` and toggled when the button is clicked.

`aria-hidden` should be applied to the `<NavMenu>` component. On desktop this will be set to false, but on mobile this should be toggled when the hamburger nav is clicked.

magnificode:
![](https://github.trello.services/images/mini-trello-icon.png) [[PrimaryNav] - Ensure proper small screen a11y](https://trello.com/c/YvFxrW56/53-primarynav-ensure-proper-small-screen-a11y)

M0nica:
I'll start working on this!

magnificode:
Thank you @M0nica !

magnificode:
Handled by #58 - closing!

PR #58, opened 2020-06-07
add additional aria-labels to PrimaryNav

M0nica:
# Describe your PR
- aria-controls  added to the <button> within the <Box> component wrapping the Hamburger menu. Points to the ID of the navigation.

- aria-expanded added to the <button> within the <Box> and toggled when the button is clicked.

- aria-hidden applied to the <NavMenu> component. On desktop set to false, but on mobile this should be toggled when the hamburger nav is clicked.

Related to # https://github.com/Rebuild-Black-Business/RBB-Website/issues/57
Fixes # https://github.com/Rebuild-Black-Business/RBB-Website/issues/57

all pages (/*) but in particular the nav

Screen Shot 2020-06-07 at 7 40 40 PM
Screen Shot 2020-06-07 at 7 40 56 PM
Screen Shot 2020-06-07 at 7 41 02 PM

## Steps to test

1. Inspect the nav element in dev tools on small and large screens and view how the DOM changes. On small screens where the hamburger menu icon appears it should also be tested by toggling the menu items.

### DIFF
--- a/src/components/PrimaryNav.js
+++ b/src/components/PrimaryNav.js
@@ -56,7 +56,7 @@ const PrimaryNav = forwardRef(
         <Box display={['block', 'block', 'none']} mr={{ sm: '5' }}>
           <button
             onClick={handleToggle}
-            aria-expanded={isVisible}
+            aria-expanded={isVisible && !isMedium}
             aria-controls="navigation"
           >
             <VisuallyHidden>
--- a/src/components/PrimaryNav.js
+++ b/src/components/PrimaryNav.js
@@ -54,7 +54,11 @@ const PrimaryNav = forwardRef(
         </Flex>
 
         <Box display={['block', 'block', 'none']} mr={{ sm: '5' }}>
-          <button onClick={handleToggle}>
+          <button
+            onClick={handleToggle}
+            aria-expanded={isVisible}
+            aria-controls="navigation"
+          >
             <VisuallyHidden>
               {`${isVisible ? 'Hide' : 'Show'} the navigation menu`}
             </VisuallyHidden>
@@ -66,6 +70,7 @@ const PrimaryNav = forwardRef(
               fill={theme.colors['rbb-white']}
             >
               <title>Menu</title>
+
               <path d="M0 3h20v2H0V3zm0 6h20v2H0V9zm0 6h20v2H0v-2z" />
             </svg>
           </button>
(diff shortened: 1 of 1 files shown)


src/components/PrimaryNav.js
```diff
@@ -54,7 +54,11 @@ const PrimaryNav = forwardRef(
         </Flex>
 
         <Box display={['block', 'block', 'none']} mr={{ sm: '5' }}>
-          <button onClick={handleToggle}>
+          <button
+            onClick={handleToggle}
+            aria-expanded={isVisible}
```
M0nica, line 59: @magnificode should this be set to true on desktop?

magnificode: @M0nica nope! Because the entire activator for the menu is hidden, this will be skipped over altogether by screen readers on desktop.

M0nica: Do you know how I can set a`aia-expanded` to`false` on desktop since `isVisible` is true on desktop?

src/components/PrimaryNav.js
```diff
@@ -78,6 +83,8 @@ const PrimaryNav = forwardRef(
           justify="flex-end"
           flexGrow={1}
           hidden={!isVisible || undefined}
+          aria-hidden={isVisible}
```
magnificode, line 86: We likely need to add a check in here to see if Hamburger menu is visible, if it's not visible (ie. on Desktop) this needs to be set to `false` to ensure that folks on larger screen sizes, using AT, know the state of the nav.

M0nica: With this current declaration `aria-hidden` defaults to`false` on desktop. Do you still think an additional check needs to be added?
Screen Shot 2020-06-07 at 7 57 03 PM

magnificode: @M0nica apologies, looking at the deploy preview, it appears that `aria-hidden` is set to `true` on desktop currently. We want to make sure this is `false` on desktop.

As soon as that medium breakpoint hits, and the navigation becomes hidden behind the hamburger menu, we want to make sure this switches to `true` until a user toggles the nav menu. Then it'll be toggled between `false` and `true` when the user hides / shows the nav.

M0nica: I realized I read that backwards 🤦🏾‍♀️. `aria-hidden` should now be false for NavMenu on desktop screens.

magnificode: @M0nica looks great! Thank you!

PR #60, opened 2020-06-08
feat: adds Helmet component with lang attribute to all pages.

chrisdemars:
# Describe your PR

This change adds the code below to all pages until we can refactor and have the `Helmet` component on one page that is generated. With this on all pages we are getting less a11y violations.

```html
<Helmet htmlAttributes={{ lang: 'en' }}>
    <title>TITLE</title>
</Helmet>
```

### DIFF
--- a/src/pages/about.js
+++ b/src/pages/about.js
@@ -3,7 +3,6 @@ import React from 'react';
 import { Flex, Heading } from '@chakra-ui/core';
 import Layout from '../components/Layout';
 import ContactCard from '../components/about/ContactCard';
-import { Helmet } from 'react-helmet';
 
 export default function About() {
   return (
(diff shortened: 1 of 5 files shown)


andrewchough:
I'm curious - for the sake of DRYing the code, can this be moved to SEO.js?

src/pages/about.js
```diff
@@ -3,10 +3,14 @@ import React from 'react';
 import { Flex, Heading } from '@chakra-ui/core';
 import Layout from '../components/Layout';
 import ContactCard from '../components/about/ContactCard';
+import { Helmet } from 'react-helmet';
 
 export default function About() {
   return (
     <Layout>
+      <Helmet htmlAttributes={{ lang: 'en' }}>
```
magnificode, line 11: @chrisdemars let's move this to the `SEO.js` component to keep things slim. This will eliminate the need for the repeated code on each page, since the `SEO.js` component is mounted on every page!

chrisdemars: @magnificode change made and pushed.

PR #62, opened 2020-06-08
Fix/add arrow functionality and adjust page neighbors

QMaximillian:
# Describe your PR

Related to # Pagination
Fixes # Adding fixes to current pagination component

Pagination Component

# Mobile 
Before - 

Screen Shot 2020-06-08 at 12 57 57 AM

Screen Shot 2020-06-08 at 12 58 20 AM

Screen Shot 2020-06-08 at 12 58 38 AM

After -

Screen Shot 2020-06-08 at 12 45 40 AM

Screen Shot 2020-06-08 at 12 47 55 AM

Screen Shot 2020-06-08 at 12 48 13 AM

#Desktop
Before - 

Screen Shot 2020-06-08 at 12 55 39 AM

Screen Shot 2020-06-08 at 12 57 05 AM

Screen Shot 2020-06-08 at 12 57 24 AM

After -

Screen Shot 2020-06-08 at 12 51 20 AM

Screen Shot 2020-06-08 at 12 54 00 AM

Screen Shot 2020-06-08 at 12 54 48 AM

## Steps to test

# Both
- Add clickable arrows to increment/decrement and show new clickable items
- First and last page are clickable at all times

# Mobile 

- List up to 5 pages before showing ellipsis
- At 5 or higher number showing, show ellipsis right before 3
- If 5 pages before final page, show those 5 pages

# Desktop

- List up 7 pages before showing ellipsis
- At 7 or higher number showing, show ellipsis right before 5
- If 7 pages before final page, show those 7 pages

### Additional notes

https://trello.com/c/XZEdvYgw/51-pagination-add-pagination-arrow-and-limit-on-desktop

https://trello.com/c/46SuY4IN/52-add-pagination-arrow-and-limits-on-mobile


.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{
```
billy-le, line 1: Can we add the `.vscode` path to the `.gitignore` file?

src/components/Pagination/Pagination.jsx
```diff
@@ -18,11 +17,12 @@ const RIGHT_PAGE = 'RIGHT';
 
 const Pagination = props => {
   const theme = useTheme();
+  const mq = window && window.matchMedia('(min-width: 480px)');
```
magnificode, line 20: We will need to do a quick check here to make sure window is not undefined before we set the variable:

```js
let mq;

if (typeof window !== 'undefined') {
  mq = window.matchMedia('(min-width: 480px)')
}
```

.vscode/settings.json
QMaximillian: I merged `upstream/master` with my `local/master` and then merged that with this branch. I checked the `gitignore` to add `.vscode` and it seems it was added!

src/components/Svgs/PaginationArrow.js
```diff
@@ -0,0 +1,54 @@
...
+      _active={{
+        color: 'white',
+      }}
+      _focus={{
+        color: 'white',
+      }}
+    >
+      <IconButton
+        onClick={onClick}
+        aria-label="Next"
```
ChrisBrownie55, line 35: This should be based on the direction the arrow is pointing. One of the two:

```js
"Go to previous page"
// or
"Go to next page"
```

And it should be applied to the `title` attribute as well for better accessibility for mouse users (hover over something with a `title` attribute and it shows the value to the user).

src/components/Svgs/PaginationArrow.js
```diff
@@ -0,0 +1,54 @@
...
+        aria-labelledby={icon}
```
ChrisBrownie55, line 36: I'm not sure this helps at all.

src/components/Svgs/PaginationArrow.js
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * @function PaginationArrow
+ *
+ * @param {function} onClick - Callback function that runs when icon is clicked
+ * @param {boolean} isDisabled - Boolean that determines whether the icon is disabled
+ * @param {string} icon - Name of the svg icon from the theme.js file
+ *
+ */
+
+const PaginationArrow = ({ onClick, isDisabled, icon }) => {
```
ChrisBrownie55, line 14: I would probably change `isDisabled` to `hidden` if you're switching over to hiding the element like the design shows. And instead of `icon` I'd suggest a `direction` prop that just takes `"left"` or `"right"`. With that you can then use the SVGs from the design easily and update the `aria-label` + `title` attributes more easily too (previous and next work as values too).

src/components/Svgs/PaginationArrow.js
```diff
@@ -0,0 +1,54 @@
...
+      _focus={{
+        color: 'white',
+      }}
+    >
+      <IconButton
+        onClick={onClick}
+        aria-label="Next"
+        aria-labelledby={icon}
+        variant="unstyled"
+        fill="white"
```
ChrisBrownie55, line 38: The `fill` needs to be based on the current state of the button:

- default: `#001514`
- active/focus: `#ffffff`
- hover: `#F46036`

I'm not 100% sure how to do that once you migrate over to actual SVGs though.

src/components/Svgs/PaginationArrow.js
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * @function PaginationArrow
+ *
+ * @param {function} onClick - Callback function that runs when icon is clicked
+ * @param {boolean} isDisabled - Boolean that determines whether the icon is disabled
+ * @param {string} icon - Name of the svg icon from the theme.js file
+ *
+ */
+
+const PaginationArrow = ({ onClick, hidden, direction = 'RIGHT' }) => {
```
ChrisBrownie55, line 14: No need for default `direction` value as it is always explicitly placed anyway.

src/components/Svgs/PaginationArrow.js
QMaximillian: The only white value in the theme is #F7F7F2. Should I inline #ffffff, create a new white value in the config, or use #F7F7F2?

Issue #63, opened 2020-06-08
Create Pagination logic


magnificode:
### Pagination

Leverage `gatsby-node.js` to create a pre-filtered page of data per each of the 50 states. This will create some nice SEO opportunities to have a URL per state of black owned businesses. It will also allow us to properly categorize data that does not have complete information, and ensure that they can still be included and searched for on the site.

jbolda:
Another option is the split it up by category, but I think that will be partly dependent on how the quantity of data of each is split up.

mbifulco:
@magnificode looks like this can be closed.  Great work @jbolda!

PR #65, opened 2020-06-08
feat: add pagination on businesses

jbolda:
# Describe your PR
This adds pagination through `gatsby-node.js`. We shifted one of the "pages" into a "template" and run `createPage` using it. This means the query needs to become a page query, and is shifted out of the Feed component. Otherwise I left the file structure as previously done the best that I could.

Related to #63 

/businesses/ will now only show 20 records (and sourced from the `const` in `gatsby-node.js`). Subsequent pages will have the page number appended to it. (Note, the default state is the filtered set based on "In Need". This will show <20 records then if some are not In Need.)

### Additional notes
I also hooked up the Pagination component, but it appears that it does not accept a currentPage prop to pass in that "state". I left that alone for now, and hopefully someone else can tackle that. (Trying to avoid getting in the weeds and keep this higher level / prototype like.)


magnificode:
@jbolda this looks great, thank you very much for knocking this out so quickly!

PR #66, opened 2020-06-08
feature/pagination-component-updates: 

billy-le:
# Describe your PR
Moved state management from useState to useReducer to reduce complexity. Removed PsuedoBox in favor of Button for accessibility reasons. Refactored code around useEffect to fix jest warnings

Related to #49 
Fixes #49

Pagination Component

image

### Additional notes
There seems to a flashing issue when clicking on the pagination and I'm not sure what may be causing it. Any advice would be greatly appreciated.

  https://www.loom.com/share/d971e3edbcd24ce3bc5d1b30b89f56b1


src/components/Pagination/Pagination.jsx
```diff
@@ -1,163 +1,220 @@
...
+ * @function range
+ *
+ * @param {number} from - Range start value
+ * @param {number} to - Range end value
+ * @param {number} step - Number to increment each loop count by
+ * @returns {Array.<number>} - An array of numbers
+ */
+
+const range = (from, to, step = 1) => {
+  let i = from;
```
ChrisBrownie55: This line is no longer needed/used per the warning.

src/components/Pagination/Pagination.jsx
```diff
@@ -1,163 +1,220 @@
...
-    }
-    if (pageNeighbors) {
-      setPageNeighbors(Math.max(0, Math.min(pageNeighbors, 2)));
-    }
+    const {
+      currentPage = 1,
+      totalRecords = 0,
+      pageLimit = 10,
+      pageNeighbors = 1,
+    } = props;
```
ChrisBrownie55: You need to destructure `props` in the component's parameter list or at the very beginning to avoid getting an error saying you haven't declared `props` as a dependency.

src/components/Pagination/Pagination.jsx
```diff
@@ -1,163 +1,220 @@
...
  *
  */
 
+function reducer(state, action) {
+  switch (action.type) {
+    case SET_PAGINATION_CONFIG:
+      return {
+        ...state,
+        ...action.payload,
+      };
```
ChrisBrownie55, line 48: The idea here isn't to have a catch-all. Rather, the syntax is:

```js
case EVENT_NAME:
  // handle specific event
```

So say you were updating the page limit based on the dependencies changing, might look like:

```js
case UPDATE_PAGE_LIMIT:
  // handle only stuff for updating the page limit
```

> Also, you're not restricted to using just `action.payload` just fyi.

src/components/Pagination/Pagination.jsx
```diff
@@ -1,163 +1,220 @@
...
+     * (1) < {4 5} [6] {7 8} > (10)
+     *
+     * (x) => terminal pages: first and last page(always visible)
+     * [x] => represents current page
+     * {...x} => represents page neighbours
      */
-    const totalNumbers = pageNeighbors * 2 + 3;
-    const totalBlocks = totalNumbers + 2;
+    const fetchPageNumbers = () => {
+      const { currentPage, pageNeighbors, totalPages } = state;
```
ChrisBrownie55: Again, you don't want to do your destructuring within the `useEffect()` because it then sees all of `state` as a dependency.

src/components/Pagination/Pagination.jsx
billy-le: In this case, should I have 5 separate dispatches for each event for the useEffect on line 78

src/components/Pagination/Pagination.jsx
billy-le: When you say beginning, do you mean at the near the top where this function component is declared?

billy-le:
@ChrisBrownie55 Could I get your review again? Your feedback has been very helpful. :)

jbolda:
@billy-le the pagination button navigates to a different page now. I believe that flash is either related to chakra styles loading slowly on the page change or a flash off "all" content followed by a re-render to the filtered state.

ChrisBrownie55:
> @ChrisBrownie55 Could I get your review again? Your feedback has been very helpful. :)

Sorry for taking so long, was stuck at work. I'll review it here quickly 👍

ChrisBrownie55:
Actually I need to change the colors for hover really quick (they don't match the ones in Figma and they're not AA compliant)

ChrisBrownie55:
Fixed! Ready to merge 👍

src/pages/allies.js
```diff
@@ -35,14 +35,8 @@ export default function Allies() {
...
         <AllyFilter onSearch={setAllyFilters} />
-        <AllyFeed filters={allyFilters} />
-        <Pagination
-          onPageChanged={pagination => {
-            // @TODO add pagination handler
-          }}
-          totalRecords={70}
-          pageLimit={5}
-        />
+        <AllyFeed />
```
magnificode, line 38: @ChrisBrownie55 ensure `filters={allyFilters}` is applied on the AllyFeed! This is the cause of the build failing here :)

PR #68, opened 2020-06-08
Add Page Hero

mbifulco:
# Describe your PR
This PR adds a component called `<PageHero />` which is used to display a title, subtitle, and header image at the top of a given page, based on the headers from [these](https://www.figma.com/file/Lo1ZcBOO9t4QRbX2I0PwBQ/RBB?node-id=933%3A6032) [mockups](https://www.figma.com/file/Lo1ZcBOO9t4QRbX2I0PwBQ/RBB?node-id=0%3A1)

Related to #n/a
Fixes #n/a

`/businesses`, `/allies`, possibly more

This is a WIP feature, I'll update pictures in conversation below as I continue work.  For now this is where I'm at:

image

## Steps to test

1. open the deploy preview
1. click `Businesses` and check it out at full size and responsive size

### Additional notes

Some other things I think we might need to tackle here:
- [x] a Heading component - this style will be shared amongst a bunch of pages, and in the spirit of DRY, should be split out into its own component
- [x] a subtitle component, for the same reasons


src/components/PageHero.js
```diff
@@ -0,0 +1,43 @@
...
+    <>
+      {heroImageUrl && (
+        <Box
+          position="absolute"
+          zIndex={-9}
+          top={0}
+          width="100%"
+          height="100vh"
+          backgroundRepeat="no-repeat"
+          background={`${
```
magnificode, line 21: Can we add a `backgroundSize="cover"` on this sucker?

src/components/Headings/PageSubtitle.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+import { Heading } from '@chakra-ui/core';
+
+const PageSubtitle = ({ children, ...rest }) => {
+  return (
+    <Heading as="h2" size="16pt" maxWidth="960px" {...rest}>
```
magnificode, line 7: Can we leverage px or rem values for the font size here please :) just for consistency!

src/components/PageHero.js
```diff
@@ -0,0 +1,43 @@
...
+        />
+      )}
+      <Box as="header" position="relative" color={theme.colors['rbb-white']}>
+        <Stack
+          zIndex={2}
+          display="block"
+          flexDirection="column"
+          alignItems="center"
+          padding="70px 24px 1.5rem"
+          minHeight="15rem"
```
magnificode, line 33: px value for height if you don't mind!

mbifulco:
One note - the business hero subtitle should have a link in it on the subscribe call to action. Do we have that URL yet?

magnificode:
> One note - the business hero subtitle should have a link in it on the subscribe call to action. Do we have that URL yet?

Not yet, that'll end up being a mailchimp form. Feel free to just hash that link for now!

PR #69, opened 2020-06-08
Feature/ally cta injection

magnificode:
# Describe your PR

Injects a Card component into the 5th position in the AllyFeed. This CTA triggers a modal that will prompt users to sign up as an Ally. The modal content will be left empty for now.

/ally

Old feed: https://share.getcloudapp.com/eDu1vnnZ
New feed: https://share.getcloudapp.com/2Nu5R6bQ

## Steps to test

1. Navigate to the /allies page
2. Ensure a stylized CTA appears in the 5th position
3. Ensure on click of the button within the CTA, a modal appears.

### DIFF
--- a/src/components/Feeds/AllyFeed.js
+++ b/src/components/Feeds/AllyFeed.js
@@ -57,63 +57,73 @@ const AllyFeed = data => {
             // @TODO :: Need to add ither cloundiary image, or background image to this card, as per the designs.
             if (index === 4)
               return (
-                <CardWrapper
-                  gridColumn={[null, null, 'span 2']}
-                  pr={theme.spacing.lg}
-                  pos="relative"
-                >
-                  <Image
-                    cloudName="rebuild-black-business"
-                    publicId="assets/ally-sign-up"
-                    transforms={{
-                      fetchFormat: 'auto',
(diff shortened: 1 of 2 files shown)


src/components/Feeds/AllyFeed.js
```diff
@@ -15,23 +49,91 @@ const AllyFeed = data => {
...
-              key={index}
-              name={allies.data.Name}
-              email={allies.data.Email}
-              specialty={allies.data.Speciality}
-              location={allies.data.Zip_Code}
-            />
-          ))}
+          {allAllies.map((allies, index) => {
+            // @TODO :: Need to add ither cloundiary image, or background image to this card, as per the designs.
+            if (index === 4)
```
racedale, line 58: Seems this would be replacing the 4th ally, rather than injecting into the 4th spot and making the list grow by one

magnificode: That is a very good point sir. If index is 4, I still need to output the ally card, in addition to this CTA. I'll get this fixed!

magnificode: @racedale should be fixed!

PR #163, opened 2020-06-10
Apply Hover effect on contact cards

brewswain:
As seen in [Trello](https://trello.com/c/gv4kT21O/116-about-page-contact-cards-need-a-hover-and-focus-state) (It wasn't opened in github issues), it was requested that ContactCard components in the /about page received a hover state.

Upon discussion in Discord, i was pointed towards doing a small scale to the actual components as it would be in line with the motif used for our social links.

The procedure I used was to use a PseudoBox and wrap the Card components with it as it has access to hover and focus states:
image

Please see below for how I implemented them. 

image

One potential issue in my approach was that to apply these styles I created a style object here:

image

I know we have a theme.js somewhere but I wasn't too comfortable adding to that file without some overview, so if you'd rather i used a global styling methodology, I'd be happy to shift around.

Another concern to this approach is scaling as, even though this works fine since we only have 3 ContactCards for now, if we were to want to use hover state on a bunch of dynamically generated cards,, this could become a bit of a nightmare. 

However, especially considering how close we are to launch, it works fine within our current needs:

[screencast here!](https://d.pr/v/Rd1lOi)


magnificode:
@brewswain - no worries on adding stuff to the theme file, appreciate you being cautious there! Since we're trying to power through as much as we can I'll approve this and we can refactor some values out into themes down the road. Thank you!

PR #169, opened 2020-06-11
Truncate description to 250 characters

mbifulco:
# Describe your PR

Fixes https://trello.com/c/Auq603FR/161-business-result-card-blurb-should-be-cut-off-at-100-char

The trello card called for 100characters, but that ends up being very short.  I'm suggesting 250 instead.

100 char
image

250 char
image

## Steps to test

1. check out `/businesses` - there shouldn't be any _crazy long_ entries any longer!

### Additional notes
Eventually this should be extended by providing a modal that shows the entire card with description if the length is longer than the cutoff

### DIFF
--- a/src/components/ResultCard.js
+++ b/src/components/ResultCard.js
@@ -2,6 +2,7 @@ import { Box, Heading, Icon, Text, useTheme } from '@chakra-ui/core';
 import PropTypes from 'prop-types';
 import React, { forwardRef } from 'react';
 import Link from '../components/Link';
+
 import { zipcodeConversion } from '../utils/locationUtils';
 import { toCamelCase } from '../utils/stringUtils';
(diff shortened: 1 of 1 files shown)


billy-le:
have you looked into using css `line-clamp` here to truncate the lines to three lines? In my experience, truncating character count can lead to inconsistencies to the layout if the font is not mono spaced.

https://css-tricks.com/almanac/properties/l/line-clamp/

So far, IE and Opera Mini doesn't support it.
https://caniuse.com/#search=line-clamp

magnificode:
@billy-le that's a great suggestion. We're in merge what we can super quickly mode at the moment so I'm going to merge this for now. If we're able put up another PR that leverages line clamp then by all means!

PR #173, opened 2020-06-11
Feature/404 page

juanri0s:
# Describe your PR
Add 404 page

https://www.figma.com/file/Lo1ZcBOO9t4QRbX2I0PwBQ/RBB?node-id=1550%3A8437

Related to #
Fixes #

/400

Desktop:
image

Mobile (image should be between text but couldn't get it)
image

## Steps to test

1. Go to /400

### DIFF
--- a/src/pages/404.js
+++ b/src/pages/404.js
@@ -6,7 +6,7 @@ import Image from '../components/Image';
 import Layout from '../components/Layout';
 import Link from '../components/Link';
 
-export default function About() {
+export default function NotFound() {
   const theme = useTheme();
   return (
     <Layout>
--- a/src/pages/404.js
+++ b/src/pages/404.js
@@ -1,0 +1,133 @@
+import { Flex, List, ListItem, Text, useTheme } from '@chakra-ui/core';
+import { graphql, StaticQuery } from 'gatsby';
+import React from 'react';
+import ErrorBoundary from '../components/ErrorBoundary';
+import Image from '../components/Image';
+import Layout from '../components/Layout';
+import Link from '../components/Link';
+
+export default function About() {
+  const theme = useTheme();
+  return (
+    <Layout>
+      <ErrorBoundary>
+        <StaticQuery
+          query={NotFoundLinks}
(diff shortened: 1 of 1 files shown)


juanri0s:
[Make 404 page](https://trello.com/c/RD6ZIpXS/166-make-404-page)

src/pages/404.js
```diff
@@ -0,0 +1,133 @@
...
+                  <ListItem>
+                    <Link variant="standard" to="/allies" color="#000">
+                      View Allies
+                    </Link>
+                  </ListItem>
+
+                  <ListItem>
+                    <Link
+                      variant="standard"
+                      to="http://join.rebuildblackbusiness.com/"
```
jordanandree, line 87: I recently added to this URL to [src/constants/about.js](https://github.com/Rebuild-Black-Business/RBB-Website/blob/master/src/constants/about.js). Since its pretty global, maybe its better suited now in [gatsbsy-config.js](https://github.com/Rebuild-Black-Business/RBB-Website/blob/master/gatsby-config.js) as a new key / value in the `module.exports` object? This can probably be separate PR too if its too far out of scope.

src/pages/404.js
```diff
@@ -0,0 +1,133 @@
...
+                  as="h1"
+                  fontFamily={theme.fonts['heading-slab']}
+                  color={theme.colors['rbb-gray']}
+                  fontWeight="900"
+                  fontSize="64px"
+                  lineHeight="77px"
+                  textTransform="uppercase"
+                  marginBottom="0.625rem"
+                >
+                  OUR BAD,
```
jordanandree, line 36: I see its in the figma, but the comma at the end feels a little off and like it shouldn't be there. Not a deal breaker at all but just thought I'd point it out.

juanri0s: I agree. Will bring it up to design team

src/pages/404.js
juanri0s: yupp config would be best, thank you for mentioning it

PR #198, opened 2020-06-11
Swap Google Tag Manager for Google Analytics

mbifulco:
# Describe your PR
Removes the plugin for google tagmanager and includes one for google analytics instead. We may choose to swap back at some point, but for now this is simpler.

**Note:** this requires changing an environment variable - it used to be called `GOOGLE_TAGMANAGER_ID` and is now `GOOGLE_ANALYTICS_PROPERTY_ID`.  It should have the same value (`UA-somethingsomething`), but there's an important difference between the two ID types, which I didn't want to confuse.

## Steps to test

1. update that environment variable, then redeploy this branch and check google analytics for live activity

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -175,13 +175,14 @@ module.exports = {
       },
     },
     {
-      resolve: 'gatsby-plugin-google-tagmanager',
+      resolve: `gatsby-plugin-google-analytics`,
       options: {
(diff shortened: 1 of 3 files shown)


magnificode:
@mbifulco I've updated the env variable on Netlify, thank you for this!

mbifulco:
Awesome! I just confirmed that it's working on the netlify preview, too.  💯

PR #201, opened 2020-06-11
Fix/remove target="_blank" from allies page

kieranklaassen:


shapirone:
My 2 cents - I don't think we should merge this.

This PR was requested in the name of A11y, but it is not a WCAG AA guideline to not use `_blank`. They do make the recommendation to only use it when necessary. In this instance, we are taking a user away from the page they want to be on to view a legal doc, which is very disruptive to the flow.

Ref articles:
https://www.w3.org/TR/WCAG20-TECHS/G200.html
https://codersblock.com/blog/external-links-new-tabs-and-accessibility/

magnificode:
> My 2 cents - I don't think we should merge this.
> 
> This PR was requested in the name of A11y, but it is not a WCAG AA guideline to not use `_blank`. They do make the recommendation to only use it when necessary. In this instance, we are taking a user away from the page they want to be on to view a legal doc, which is very disruptive to the flow.
> 
> Ref articles:
> https://www.w3.org/TR/WCAG20-TECHS/G200.html
> https://codersblock.com/blog/external-links-new-tabs-and-accessibility/

Whoops, somehow missed this. You're absolutely right. I think we'll forego this change in favor of maintaining the flow of the user. I think we can equate this to a 'cart' type of process as well. If someone is coming to the allies page and searching for an ally, we don't want to undo their search query if they switch over to the legal page.

Appreciate the thought, we definitely want to make sure we're very deliberate when opening new tabs, this is one of those cases :)

PR #207, opened 2020-06-11
Algolia page numbers are 0 indexed, this caters for that.

llexical:
So yeh. As the title says. Algolias page numbers are 0 indexed. No joke.
Page 1 = page0. Kinda like our current project versioning xD.

This makes sure that we always pass page -1 to aloglia for searching so
we never miss out on any results. I am hoping this will solve much
oddness.


racedale:
This is gonna become a crucial test we should write after launch

magnificode:
> This is gonna become a crucial test we should write after launch

💯 - can we write out an issue in GH / Trello to track this?

PR #208, opened 2020-06-11
Fix/hero contrast

subject026:
To Improve the contrast between text and hero bg image/gradient I've increased the default height of the hero on smaller screens and tweaked the gradient a little. 

Also added a prop so that the default height can be overridden. The hero is used on the businesses page and the hero needs to be a bit taller so that the form labels are visible.

Related to https://trello.com/c/TZupYF0t
Improves contrast between text and hero bg image/gradient.

PageHero component, used on allies and businesses pages

## Before

hero-contrast_before
hero-contrast-businesspage

## After
hero-contrast_after
hero-contrast-businesspage_after


subject026:
looking at this again I think the contrast on the business page should be improved a bit more

magnificode:
@subject026 - yeah this is a tough one...on super narrow screens we still see the issue: https://share.getcloudapp.com/KouBQkOY

I think we need a bit of an aggressive solution here to make sure on devices of all shapes and sizes, now and in the future, never have contrast issues with the labels.

magnificode:
@subject026 - Just had a chat with design, here's what we'd like for now:

At our smallest breakpoint we'd like to add a background color on the container housing the filters. If we could use the same color that's being used as the background color in the nav that would be great. We'll need to change the input label color to black at this breakpoint as well.

Ensure there's a comfortable bit of padding around the filters on these smaller screens as well.

Let me know if you have any questions!

subject026:
Background for the filter forms has now been added for small screens and I've changed the label text colour. Reversed the changes I made to the hero/gradient too.

There are identical styles hard coded across the 2 filter components atm... could do with creating some reusable form/label components at some point. 

filters_new

Issue #211, opened 2020-06-11
Hide pagination when no business are matched in search


devmrin:
On the `/businesses` page upon search if no businesses are matched - the pagination should be hidden. 

Screenshot for reference: 
![image](https://user-images.githubusercontent.com/26342387/84420961-79cf5b80-ac38-11ea-893a-ce06df6c5936.png)

juanri0s:
![](https://github.trello.services/images/mini-trello-icon.png) [On Business Page we need to hide pagination when no business are matched in search.](https://trello.com/c/WDpG3UVE/202-on-business-page-we-need-to-hide-pagination-when-no-business-are-matched-in-search)

PR #214, opened 2020-06-11
removed pagination if no results

molebox:
This PR removes the paginations links if no results are shown. I was unable to get no results as the search broke, so i couldnt take a screenshot. But it works, trust me :-)

Businesses

## Steps to test

Run locally and try and get no results


.env.development.EXAMPLE
```diff
@@ -1,3 +0,0 @@
-GATSBY_ALGOLIA_APPLICATION_ID=""
-GATSBY_ALGOLIA_SEARCH_KEY=""
-GATSBY_ALGOLIA_INDEX="businesses"
```
magnificode, line 3: @molebox looks like this may have been accidentally removed!

molebox: oh sorry, i just renamed it. Can add back

juanri0s:
[On Business Page we need to hide pagination when no business are matched in search.](https://trello.com/c/WDpG3UVE/202-on-business-page-we-need-to-hide-pagination-when-no-business-are-matched-in-search)

ChrisBrownie55:
Looks like this fixes #211. Closing now.

PR #219, opened 2020-06-11
fixed sidenav scroll issue

molebox:
This PR fixes the mobile nav scrolling issue, whereby the user could scroll the page when the nav is open.

Ive left some comments in the code so its clear whats going on. Not super complicated or anything, just so people know

All pages

Couldnt really take a screenshot of this behavior 

## Steps to test

Local dev, dev tools, mobile view. Then click the nav, it should lock in place, navigate to a new page and you should be able to scroll. Rinse and repeat


gatsby-browser.js
```diff
@@ -1 +1,9 @@
 import '@reach/skip-nav/styles.css';
+
+// On the route update we can access the current path and add a state to each navigation, which we call prevLocation.
+// Defaults to the home route as thats where we start
+export const onRouteUpdate = ({ location, prevLocation }) => {
+  if (location && location.state) {
+    location.state.referrer = prevLocation ? prevLocation.pathname : '/';
```
magnificode, line 7: @molebox ah, yeah we're going to need to refactor this a bit. Even on a fresh page load `referrer` does not exist. Defaulting to 'null' is likely the right move here.

molebox: Yep, good shout, it crossed my mind but I chose (incorrectly) to not go that route. I can fix it tomorrow if there is no rush, otherwise it's a pretty simple fix.

src/components/PrimaryNav.js
```diff
@@ -28,12 +29,31 @@ const PrimaryNav = forwardRef(
...
+    useEffect(() => {
+      // if the sidenav is open and the pages path does not match the previous path then we unset the position: fixed on the body
+      if (
+        !isVisible &&
+        location.state !== null &&
+        location.pathname !== location.state.referrer
+      ) {
+        document.body.style.position = 'unset';
+      }
+    }, [isVisible]);
```
mbifulco, line 55: These may not change, but should be included in the dependency array here, I believe

```suggestion
    }, [isVisible, location.pathname, location.state]);
```

src/components/PrimaryNav.js
```diff
@@ -28,12 +29,31 @@ const PrimaryNav = forwardRef(
...
+      document.body.style.position = 'fixed';
       setIsVisible(!isVisible);
     };
 
+    useEffect(() => {
+      // Handle the first render when the location state will be null.
+      if (!isVisible && location.state === null) {
+        document.body.style.position = 'unset';
+      }
+    }, []);
```
mbifulco, line 44: Since this is meant to run on first render,  we should be able to disable the eslint rule here

```suggestion
	/* eslint-disable react-hooks/exhaustive-deps */
    useEffect(() => {
      // Handle the first render when the location state will be null.
      if (!isVisible && location.state === null) {
        document.body.style.position = 'unset';
      }
    }, []);
    /* eslint-enable react-hooks/exhaustive-deps */
```

PR #220, opened 2020-06-11
remove duplicate close button from modals

M0nica:
# Describe your PR

Related to # https://trello.com/c/PLqBskue/208-remove-large-close-button-on-submit-business-form
Fixes #

It looks like that's where the "submit" button would typically go, users might click it erroneously

Users can already hit 'x' or tap/click outside the modal to dismis

Business Form Modal and other models that had both an 'x' and close button

Before
image

After
Screen Shot 2020-06-11 at 5 44 21 PM

## Steps to test

1. Go to /businesses
2. Click "Register a business" button
3. Click 'X' or outside of modal to close modal and no longer see  the close button


magnificode:
@M0nica Ah, since we're no longer leveraging the button in the footer we can remove the import for `ModalFooter`! Hooray for deleting code 😄 - I'm going to just leave this one comment here, however it applies to all files in the PR.

It looks like there's a couple files that are no longer using `Button` as well!

PR #232, opened 2020-06-12
[a11y] Unset footer Navmenu from being Flex, and add explicit flex styling

### LINKED ISSUE #230: Fix uncontained <li> in Footer for a11y
Resolve issue from Lighthouse scan:
List items (<li>) are not contained within <ul> or <ol> parent elements.

racedale:
# Describe your PR
Fixes a11y issue for Lighthouse that <li> need to be contained in UL or OL tag.

Fixes #230 

Footer

No visual difference

## Steps to test

1. Inspect Element on footer to see <ul> tag around nav links


src/components/footer/FooterLinks.js
```diff
@@ -13,11 +12,11 @@ const FooterLinks = () => {
         render={data => {
           return (
             <NavMenu
-              as={Flex}
-              w={['90%', '90%', '90%', 'full']}
-              justify="space-evenly"
-              wrap={['wrap', 'wrap', 'wrap', null]}
-              align="center"
+              style={{
```
magnificode, line 15: @racedale did you try doing `as={List}` instead of `as={Flex}`

At that point we should be able to do `justifyContent='space-evenly'` and `w="100%"` as props.

I think. Still not 100% solid with my Chakra skills.

racedale: Yeah tried that, `NavMenu` is a List component by default which is why I didn't do `as={List}`, but it [doesn't support any of the flex box props](https://chakra-ui.com/list#props) which makes sense since it's not the Flex component anymore

PR #238, opened 2020-06-12
Wait for loaded state before rendering business filter

jordanandree:
this fixes a bug where the filters don't render the preset filters passed through URL params.

/businesses page

|Before|After|
|---|---|
|image|image|

## Steps to test

_Before_
1. Got to /businesses
2. Set the Business Type to something other than All
3. Reload the page
4. Business Type is set to All when it should be the filter value

_After_
1. Got to /businesses
2. Set the Business Type to something other than All
3. Reload the page
4. Business Type is set to the filter value


jordanandree:
[URL params don't apply to business filter form state](https://trello.com/c/ETAUWXEX/225-url-params-dont-apply-to-business-filter-form-state)

PR #242, opened 2020-06-12
add eslint precommit hook

jamiebuilds:
# Describe your PR

This will run eslint as an additional pre-commit hook.

None

## Steps to test

1. Make a file change with a lint error to a `.js` file
2. Run `npx lint-staged`

### Additional notes

@racedale asked me to make this change

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,6 +89,7 @@
     }
   },
   "lint-staged": {
-    "*.{js,jsx,ts,tsx,d.ts,css,md}": "prettier --write"
+    "*.{js,jsx,ts,tsx,d.ts,css,md}": "prettier --write",
+    "*.{js,jsx,ts,tsx}": "eslint"
   }
 }


package.json
```diff
@@ -89,6 +89,7 @@
     }
   },
   "lint-staged": {
-    "*.{js,jsx,ts,tsx,d.ts,css,md}": "prettier --write"
+    "*.{js,jsx,ts,tsx,d.ts,css,md}": "prettier --write",
+    "*.{js,jsx,ts,tsx}": "eslint"
```
racedale, line 93: We probably want this as `eslint --fix` to auto fix easy things

jamiebuilds: I've done that in the past and I ended up removing it.

ESLint and Prettier have a different bar when it comes to what is allowed to be autofixed. Prettier will never ever make changes to how your code functions when making autofixes. ESLint plugins will sometimes make unsafe changes in autofixes, and I've seen that cause confusion.

I can definitely change it though if you'd be okay with that risk

racedale: yeah that's reasonable!

PR #245, opened 2020-06-12
Fix/Pagination hover effect sometimes persistent

ajadams:
# Describe your PR
The ticket states:
> Chrome, desktop and mobile - at times, the hover effect on pagination stays even after cursor no longer hovered / after click.

The only way I was able to reproduce was clicking on a link then tabbing (see `Steps to test` for more deets) to have the effect persist. But it makes sense that this would be a problem when you look at [these two lines](https://github.com/Rebuild-Black-Business/RBB-Website/blob/446b87b2781f81155886e617eefaf1fdcfc9edae/src/components/Pagination/Pagination.jsx#L148-L149) - it's very possible for both the hover and focus state to occur simultaneously. So what I ended up doing is storing refs to each pagination page button element to handle blurring the currently focused ref when user starts hovering to prevent simultaneous active states for hover and focus.

Ticket: https://trello.com/c/zxapOTEw/139-pagination-hover-effect-sometimes-persistent

This affects the pagination page links, both the hover and focus states.

| Before (buggy) | After (fix) |
|-|-|
|[screencap from ticket](https://trello-attachments.s3.amazonaws.com/5ed7c609e0fa8c77ca90e9ff/5ee148f4281b312695f94401/f5d87ca253af89a3fd3309ae4181759e/rbb_pagination.gif)| [screencap from local](https://share.getcloudapp.com/6qu2LkA2)|

## Steps to test
*To recreate the bug*:
1. Go to /businesses
2. Scroll down to pagination element
3. click a page (that will set focus to that elem
4. spam tab key to tab through the pages (tab will set focus to the <a> then the <button> that's why you have to press twice in order to go to the next page number). So press tab until the <button> has focus and it'll have the gray bg color.
5. Try hovering over any other pages, the focused button will still have gray bg while the currently hovered item will _also_ have a gray bg.

For the fix - do the same steps above^ , except this time, the focus is removed from the element when you start hovering.

### Additional notes
Not gonna lie, it's been a minute since I've done some more complex react work, so I relied on [this guide](https://erikmartinjordan.com/multiple-refs-array-hooks) for this solution 😅

### DIFF
--- a/src/components/Pagination/Pagination.jsx
+++ b/src/components/Pagination/Pagination.jsx
@@ -92,12 +92,31 @@ function Pagination({ location, currentPage, totalPages }) {
     [...new Array(pages.length)].map(() => React.createRef())
   );
   const [refWithFocus, setRefWithFocus] = useState();
+  const [refWithHover, setRefWithHover] = useState();
 
-  // Blurres currently focused elem and then resets the state
-  function removeFocus() {
+  /**
+   * Sets state of the ref that currently is being hovered and unset state of refWithFocus to remove the focus styles when elem is being hovered
+   *
+   * @param {HTMLElement} elem
+   */
+  function addHoverRemoveFocus(elem) {
     if (refWithFocus) {
       refWithFocus.blur();
+      setRefWithHover(elem);
       setRefWithFocus(null);
+    }
+  }
+
+  /**
+   * Sets state of the ref that currently has focus and unset state of refWithHover to remove the hover styles when elem is being focused
+   *
+   * @param {HTMLElement} elem
+   */
+  function addFocusRemoveHover(elem) {
+    setRefWithFocus(elem);
+
+    if (refWithHover) {
+      setRefWithHover(null);
     }
   }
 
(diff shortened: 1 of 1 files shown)


src/components/Pagination/Pagination.jsx
```diff
@@ -1,4 +1,4 @@
-import React, { useMemo } from 'react';
+import React, { useMemo, useRef, useState, useEffect } from 'react';
```
mbifulco, line 1: Hey @ajadams! Thanks for the PR.  It looks like `useEffect` isn't used in this file.  Can you remove it from this line?


```suggestion
import React, { useMemo, useRef, useState } from 'react';
```

racedale:
Tabbing to focus then hovering to focus seems fixed. The opposite still causes two focus states: hovering to focus then tabbing to focus

Screen Shot 2020-06-12 at 7 54 26 AM

src/components/Pagination/Pagination.jsx
mbifulco, line 1: Hey @ajadams! Thanks for the PR.  It looks like `useEffect` isn't actually used here.  Can you remove it?

```suggestion
import React, { useMemo, useRef, useState } from 'react';
```

ajadams: good catch - removed!

ajadams:
@racedale oh good call, I can work on that in a separate ticket. Or can just submit a second/follow-up PR under this same ticket and just reference this one to keep code changes small

@mbifulco thanks for cutting that a11y ticket, I can take a stab at it

ajadams:
Ended up just adding those changes to this PR, hope that's ok! The changes in 27736a0 address what you were experiencing @racedale

PR #249, opened 2020-06-12
Stop the pagination page 1 from removing location

llexical:
When you went to page 1 in the pagination it would clear all url
parameters, which was lazyness on my side when I implemented it,
didn't think of this use case.

Tgis fix ensures that we only remove the pages search param on page 1.


juanri0s:
[When changing search type to all from in need with predefined search parameters it just returns all results.](https://trello.com/c/QNWA1y7N/205-when-changing-search-type-to-all-from-in-need-with-predefined-search-parameters-it-just-returns-all-results)

PR #246, opened 2020-06-12
Prevent additional algolia requests by circumventing gatsby `navigate`

jordanandree:
Using the business filters add an additional XHR request with empty coordinates. This has potential to cause a race condition with rendering where location-filtered results will show and then unfiltered results will jump into view.

There is some nuance to this that I don't fully understand, but it something to do with how `navigate` is causing a full `Business` component re-render.

## Pages/Interfaces that will change

/businesses

### Screenshots / video of changes

Before video of additional XHR requests: https://share.getcloudapp.com/6qu2LWJR
After video: https://share.getcloudapp.com/2Nu5r2Ro

## Steps to test

- Go to businesses page
- Search for Portland and submit
- Change "show me" to all and submit

PR #250, opened 2020-06-12
Move from navigate to window.locate for search filters

llexical:
Navigate in Gatsby was causing a full rerender of the page on trigger,
which meant that n1, you were pushed to the top of the page quite
violently every time you did a search but also with a search being
programatically triggered, it would then get retriggered by the page
refresh. There would then be a race condition and everything got a bit
strange.

One way to solve this would have been to stuck with just using page
refreshes however I much prefer a dynamic search without jumping around
all over the place so went in that direction.

I have stored a page location object which I updated with the new path
name and search params when filters change the URL which allows it to
act like navigate for the pagination which uses page location. I could
have setup pagination to then change page dynamically as well but then
you need to scroll to things and page refreshing is an easy way to skip
all that. Also means the pagination would not have been as easily
reusable.

Its a bit of a Frakensteins monster, however I hope it makes sense!
Happy for any review comments/ideas.

# Describe your PR

Related to #
Fixes #

## Steps to test

1.

### DIFF
--- a/src/hooks/usePagination.js
+++ b/src/hooks/usePagination.js
@@ -6,6 +6,8 @@ function usePagination(location, onChange) {
   useEffect(() => {
     const pageParam = new URLSearchParams(location.search).get('page');
 
+    // Duplicated because otherwise it wants it to be inside useEffect
+    // And then that gets complex. Easier this way for now!
     if (pageParam) {
       setPage(pageParam);
       onChange(pageParam);
@@ -15,7 +17,7 @@ function usePagination(location, onChange) {
     }
   }, [location.search, onChange]);
 
-  return page;
+  return { page };
 }
 
 export default usePagination;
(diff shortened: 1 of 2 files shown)


llexical:
Forgot to say, much credit to: https://github.com/Rebuild-Black-Business/RBB-Website/pull/246 who figured out what was going on!

mbifulco:
@llexical how can I test this out?

racedale:
@mbifulco Here's the trello card with details
https://trello.com/c/QNWA1y7N/205-when-changing-search-type-to-all-from-in-need-with-predefined-search-parameters-it-just-returns-all-results

llexical:
@mbifulco Heres a link: https://5ee38e9165cec40007cd3574--rebuild-black-business.netlify.app/

PR #254, opened 2020-06-12
Hotfix/205 when changing search type returns all results + some fixes

llexical:
Original bug: When changing search type to all from in need with predefined search parameters it just returns all results.

So the fix for the above bug was released and then reverted because it
broke pagination. I have now tested pagination to death AND fixed the bug
where multiple algolia requests were being made on page change that
created a race condition and then broke things. SO hopefully we have
fixed many things.

--- Additional Extras ----
Basically a tidying things up that were annoying me commit.

Added a loading state to the spinner, seemed to come in the UI library
so was pretty easy to add.

Fixed the no results flash so that it will only show we we have
confirmed that there really are no results.

Resized the images to an appropriate size and set DPR to 2 so they can
be retina.

Related to: https://github.com/Rebuild-Black-Business/RBB-Website/pull/253
Fixes 
- https://trello.com/c/QNWA1y7N/205-when-changing-search-type-to-all-from-in-need-with-predefined-search-parameters-it-just-returns-all-results
- https://trello.com/c/KyTSKyl7/233-sorry-no-business-matched-this-search
- https://trello.com/c/1u0qaP7O/229-business-images-in-the-business-feed-are-5-6-times-bigger-than-they-need-to-be

- Businesses page

## Steps to test

Process:
- Go to businesses page
- Search for Portland and submit
- Search for all and submit

Expected Result:
- Would show the number of pages for results in Portland

### DIFF
--- a/src/components/Feeds/BusinessFeed.js
+++ b/src/components/Feeds/BusinessFeed.js
@@ -3,30 +3,28 @@ import PropTypes from 'prop-types';
 import React, { useEffect, useState } from 'react';
 import NoResultsCard from '../Cards/NoResultsCard';
 import BusinessFilter from '../Filters/BusinessFilter';
+import { LOADING_STATE } from '../../hooks/useAlgoliaSearch';
+
 import ResultCard from '../ResultCard';
-import CardSkeleton from '../Loading/CardSkeleton';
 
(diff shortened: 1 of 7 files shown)


llexical:
Test Link: https://5ee3e52c4a6a300007021d08--rebuild-black-business.netlify.app/

PR #257, opened 2020-06-18
Feature/circle ci setup

racedale:
# Describe your PR
Sets up a basic Circle CI workflow to run Cypress integration tests to work with forked PR's. It's not required for merging yet since the tests aren't passing. (a11y and navigation tests to fix)

## Steps to test

1. Check https://app.circleci.com/pipelines/github/Rebuild-Black-Business/RBB-Website

### Additional notes
Will be changing it over to only do builds on PR's instead of any commit ever after this.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,10 +21,10 @@ workflows:
               only:
                 - staging
                 - trying
-      - publish:
-          requires:
-            - test
-          # only publish site changes on 'merge'
-          filters:
-            branches:
-              only: staging
+      # - publish:
+      #     requires:
+      #       - test
+      #     # only publish site changes on 'merge'
+      #     filters:
+      #       branches:
+      #         only: staging
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,17 +10,26 @@ jobs:
       - node/with-cache:
           steps:
             - run: npm ci
-            - run: npm run test:e2e:ci
+            - run:
+                name: 'Cypress Tests'
+                command: npm run test:e2e:ci
+                environment:
+                  AIRTABLE_API_KEY: $AIRTABLE_API_KEY
+                  AIRTABLE_BASE_ID: $AIRTABLE_BASE_ID
+                  GATSBY_ALGOLIA_APP_ID: $GATSBY_ALGOLIA_APP_ID
+                  GATSBY_ALGOLIA_SEARCH_KEY: $GATSBY_ALGOLIA_SEARCH_KEY
+                  GATSBY_ALGOLIA_INDEX: $GATSBY_ALGOLIA_INDEX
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,24 @@
...
+    executor:
+      name: node/default
+    steps:
+      - checkout
+      - run: npm ci
+      - run:
+          name: 'Cypress tests'
+          command: npm run test:e2e:ci
+          environment:
+            AIRTABLE_API_KEY: $AIRTABLE_API_KEY
```
llexical, line 15: Not sure if I am missing something but you shouldn't need to define these, if they are setup as environment variables in circle they should just work as is.

llexical: Assuming its something to do with running cypress? I've seen a lot of weird and wonderful ways of getting it working in circle lol.

racedale: Yeah we didn't need them defined here good catch!

racedale:
@llexical I agree that having failing tests is bad, but these tests are already in the repo and failing. I was wanting to get this done in multiple parts that way it's not so dependent on one person. 
Someone else could easily work on improving the tests. 

With that said though, the main reason for test failures is a11y color contrast reasons, and that's going to require feedback from design (already let them know about it).  I have found a few fixes for some of the other errors but wanted to do it as it's own PR so it's easier to review. Unless we `.skip` the a11y tests we wouldn't be able to fix it now, and I think that's worse than having failing tests because they are less likely to ever be fixed. Thoughts?

racedale:
We _could_ set the tests to disable the color contrast rule, but theres not a great way to enforce that we un-disable it in the future. We could make a GitHub issue for it but that's the best idea I can think of.  

```
{
    rules: { 'color-contrast': { enabled: false } },
}
```

llexical:
> @llexical I agree that having failing tests is bad, but these tests are already in the repo and failing. I was wanting to get this done in multiple parts that way it's not so dependent on one person.
> Someone else could easily work on improving the tests.
> 
> With that said though, the main reason for test failures is a11y color contrast reasons, and that's going to require feedback from design (already let them know about it). I have found a few fixes for some of the other errors but wanted to do it as it's own PR so it's easier to review. Unless we `.skip` the a11y tests we wouldn't be able to fix it now, and I think that's worse than having failing tests because they are less likely to ever be fixed. Thoughts?

I would rather skip than have failing tests that we ignore and merge anyway. If you have failing tests then people will ignore them and they may as well not be there, often they mean people will not check the tests to see if the ones normally passing do and so they assume and it gives false confidence. Also normalising a failing test does not make them more likely to get fixed, I've seen teams ignore them for years xD.

If we never make the tests work then thats a different issue. Either we don't care about the tests and they are not useful so we don't need them or we do and so we should prioritise fixing them. Theres no point having tests we are not committing to maintaining or we may as well not have any at all.

PR #259, opened 2020-06-19
Mobile/desktop ally and businesses "search" button text is off vertical center

ajadams:
# Describe your PR
The text alignment of the "search" text in the business and ally filter components are vertically misaligned on Safari (iOS & desktop), as specified in [the Trello ticket](https://trello.com/c/gtuMJ9t4/227-safari-mobile-ally-and-businesses-search-button-text-is-off-vertical-center). After digging in, apparently `flex` styles do not work on `button` or `fieldset` elements in Safari, as mentioned in this [stackoverflow thread](https://stackoverflow.com/a/44756423) (TIL this was a problem). Changing the `Button` component to render as an anchor tag did in fact fix the issue.

In this PR, I added `as="a"` to the `PrimaryButton` component to render a `<a type="button">` instead of `<button>` to allow `flex` layout styles to work as expected, and made the change there to prevent future uses of `PrimaryButton` elsewhere.

Afaik, the `businessFilter` and `allyFilter` are the only areas that are using a flex-styled `<button>`, which will change to a `<a type="button">`

| Before (prod on safari) | After (local on safari |
|-|-|
|https://share.getcloudapp.com/8Lu7n0Nk|https://share.getcloudapp.com/Z4uYqApD|

## Steps to test
1. Open safari
2. Navigate to /businesses or /allies
3. Locate the "search" button, above the search results
4. Can verify text is vertically aligned by comparing it to prod in Chrome

### Additional notes
- [caniuse](https://caniuse.com/#feat=flexbox) looks to be incorrectly displaying that the flex layout props should be working as expected on latest safari 😬 
- There's actually multiple ways of solving this. I went for the road of less changes 😅 , but this [readme](https://github.com/philipwalton/flexbugs#9-some-html-elements-cant-be-flex-containers) shares which elements can't use flex on which browsers. Also shares workaround for them, until a fix is distributed. I'm def not married to my attempt - I am open to other solutions!

### DIFF
--- a/src/components/Buttons/PrimaryButton.js
+++ b/src/components/Buttons/PrimaryButton.js
@@ -7,6 +7,7 @@ function PrimaryButton(props) {
   return (
     <Button
       {...props}
+      as="a"
       bg={theme.buttons.primary.backgroundColor.default}
       color={theme.buttons.primary.color.default}
       borderRadius="button"


ajadams:
Don't have CircleCI creds, so not sure what's failing exactly.
Can one of the admins let me know? 🙏

magnificode:
@ajadams - I believe the CircleCI stuff is new. I'll get with @racedale to figure out what's failing!

src/components/Buttons/PrimaryButton.js
```diff
@@ -7,6 +7,7 @@ function PrimaryButton(props) {
   return (
     <Button
       {...props}
+      as="a"
```
magnificode, line 10: Since this button is submitting the filter options, we may want to consider rendering this as an `input` with a `type="submit"` rather than an `a`. I think that should help to maintain proper syntax on this element.

magnificode:
@ajadams - thank you again for the super well laid out PR! Left one comment around the semantics of the type of element we render the button as. Happy to discuss further if you have any additional thoughts!

src/components/Buttons/PrimaryButton.js
ajadams: Good point about proper syntax - makes sense to me!

I do want to throw another option your way, that would allow to avoid changing button semantics entirely:
I was curious why [Chakra UI buttons](https://chakra-ui.com/formcontrol#usage-with-form-libraries) didn't have this Safari issue and after some digging, I think it's because we are setting [top & bottom padding](https://github.com/Rebuild-Black-Business/RBB-Website/blob/766c6c9d389a2a2f2ddfd01cdefb3a60e12dfb20/src/components/Buttons/PrimaryButton.js#L20) whereas Chakra UI is not. Adding top/bottom padding in combination with the `height` prop provided by Chakra is where Safari, I believe, is having issues properly aligning. So I am suggesting we [set top/bottom to `0`](https://github.com/Rebuild-Black-Business/RBB-Website/blob/0f09b3382104d8b9903e9084be926ccf2c78a3ef/src/gatsby-plugin-chakra-ui/theme.js#L164) for the primary button theme, and probably all of our button theme variations. This way, it's only a CSS change and the current button semantics can remain as is. Quick demo of what I'm seeing: https://share.getcloudapp.com/12uNK72J

What are your thoughts?

magnificode: @ajadams I do think we need to refactor our buttons a bit. I'd like to avoid setting a defined height on the buttons, and allow the line height / font size of the text within the button to determine the height. I'm 100% on board with a small refactor there!

Do you think it would be beneficial to close this PR out in favor of a PR more geared towards that? Happy to proceed / help however you think!

ajadams: > allow the line height / font size of the text within the button to determine the height

Totally agree with that. Ya I think closing this and cutting a refactor ticket makes sense to me!

magnificode:
Closing in favor of a Button component refactor. Thanks @ajadams !

PR #263, opened 2020-06-21
 Change over from Algolia search to PHP search backend

llexical:
We originally added in Algolia as a stop gap as we had not very many
 days before launch and needed to get something in quickly. However
 Algolia is very expensive and although the searching is great we cannot
 afford to keep running the site on it.

 Here is where @ChrisHolland came in :D. There is now a PHP backend
 setup that is super fast and works the same as Algolia for all intents
 and purposes for our search! The new useSearch hook uses that and
 removes the need for Algolia entirely from the site.

NOTE: I believe the new backend only indexes approved Businesses, so I could remove that search filter.

- Businesses page search functionality has changed but should look exactly the same.

## Steps to test

1. Go to the businesses page
2. View results
3. Search for as many different variations you can think of.
4. Go through many pages in pagination.
5. Try and break it.

## Additional Notes
- We will need to add `SEARCH_API_ENDPOINT` environment variable to Netlify.


package.json
```diff
@@ -43,6 +43,7 @@
     "gatsby-plugin-sitemap": "^2.4.5",
     "gatsby-source-airtable": "^2.1.1",
     "gatsby-source-filesystem": "^2.3.10",
+    "isomorphic-fetch": "^2.2.1",
```
magnificode, line 46: Since we're leveraging the API can we remove any of the algolia packages from here? I believe it's just the one on line 29 :)

magnificode:
Oh! @llexical any chance we can also remove the "Search powered by Algolia" bit in this PR as well?

.env.development.EXAMPLE
```diff
@@ -1,3 +1 @@
-GATSBY_ALGOLIA_APP_ID=""
-GATSBY_ALGOLIA_SEARCH_KEY=""
-GATSBY_ALGOLIA_INDEX="businesses"
\ No newline at end of file
+SEARCH_API_ENDPOINT=''
```
magnificode, line 1: This may be me doing something incorrectly but I _think_ that the env variables used on the frontend need to be prefixed with `GATSBY_` (similar to how we had the Algolia env vars set up).

llexical: Oh duh! I totally forgot that! I'll do that in a bit!

package.json
llexical: Totally forgot about that, will do that!

src/hooks/useSearch.js
```diff
@@ -0,0 +1,73 @@
...
+
+  useEffect(() => {
+    async function getBusinesses() {
+      try {
+        const filterStr = createFilterString(defaultFilters, filters);
+        const response = await fetch(
+          `${endpoint}api/v1/businesses?page=${page}&${filterStr}`
+        );
+        const results = await response.json();
+        console.log('results', results);
```
llexical, line 51: Remove this console log!

PR #272, opened 2020-06-23
Create Netlify Function to map Sentry webhook to Discord

racedale:
# Describe your PR
Manually maps Sentry webhook fields to Discord format with a Netlify Function 
similar to https://github.com/IanMitchell/sentry-discord which is setup with a Zeit/Vercel function

### Additional notes
After this is deployed, still need to setup in Sentry before it can be tested.

### DIFF
--- a/functions/sentry-discord.js
+++ b/functions/sentry-discord.js
@@ -12,7 +12,7 @@ exports.handler = function (event, context, callback) {
 
     const payload = {
       username: 'Sentry',
-      avatar_url: `https://raw.githubusercontent.com/IanMitchell/sentry-discord/master/sentry-icon.png`,
+      avatar_url: `https://sentry-brand.storage.googleapis.com/sentry-glyph-black.png`,
       embeds: [
         {
           title: body.project_name,
@@ -21,10 +21,6 @@ exports.handler = function (event, context, callback) {
           url: body.url,
           timestamp: new Date(body.event.received * 1000).toISOString(),
           color: COLORS[body.level] || COLORS.error,
-          footer: {
-            icon_url: 'https://github.com/fluidicon.png',
-            text: 'ianmitchell/sentry-discord',
-          },
           fields: [],
         },
       ],
--- a/functions/sentry-discord.js
+++ b/functions/sentry-discord.js
@@ -1,0 +1,60 @@
+const COLORS = {
+  debug: parseInt('fbe14f', 16),
+  info: parseInt('2788ce', 16),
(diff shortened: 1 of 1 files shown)


functions/sentry-discord.js
```diff
@@ -0,0 +1,60 @@
...
+  fatal: parseInt('d20f2a', 16),
+};
+
+exports.handler = function (event, context, callback) {
+  try {
+    const { body } = event;
+
+    const payload = {
+      username: 'Sentry',
+      avatar_url: `https://raw.githubusercontent.com/IanMitchell/sentry-discord/master/sentry-icon.png`,
```
M0nica, line 15: Can this be updated to not be a hot-link? i.e, link to an image hosted within this repo. I'm assuming if that image is moved or renamed it would break the avatar_url.

functions/sentry-discord.js
```diff
@@ -0,0 +1,60 @@
...
+        {
+          title: body.project_name,
+          type: 'rich',
+          description: body.message,
+          url: body.url,
+          timestamp: new Date(body.event.received * 1000).toISOString(),
+          color: COLORS[body.level] || COLORS.error,
+          footer: {
+            icon_url: 'https://github.com/fluidicon.png',
+            text: 'ianmitchell/sentry-discord',
```
M0nica, line 26: Should this be updated to point to Rebuild-Black-Business/RBB-Website?

functions/sentry-discord.js
```diff
@@ -0,0 +1,60 @@
...
+      body.event.tags.forEach(([key, value]) => {
+        payload.embeds[0].fields.push({
+          name: key,
+          value,
+          inline: true,
+        });
+      });
+    }
+
+    fetch(process.env.SENTRY_DISCORD_WEBHOOK_URL, {
```
M0nica, line 50: If this ENV variable is required to run the app locally can you update the README to let contributors know where to find or request access to this URL?

racedale: It's not required to run the app locally, it's something that will only live in the Netlify account

functions/sentry-discord.js
racedale: Actually I think better to just remove the footer altogether, doesn't really provide value here

functions/sentry-discord.js
racedale: I've updated it to point to Sentry's actual logo url from their branding page. I don't think we want to store other company's logos in our repo though

PR #273, opened 2020-06-24
Update README API Documentation link location

ifyoumakeit:
The API Documentation link is under the help button not the user image.

README section for AirTable test DB setup

### Additional notes

The https://airtable.com/appFoFzjMcciPUgoK/api/docs#javascript/authentication link referenced above this change may also have changed or be user specific because it 404s. Didn't want to update it if that link was role-specific (mine is https://airtable.com/appkenjGlBB01wr3i/api/docs#javascript/authentication). My dummy DB is **TEST DATA :: RBB :: TEST DATA**.

re: 
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ Upon clicking the above link, you will get "Read Only" access to the database.
 
 Once you've obtained access, go to the API docs and grab your API key by visiting the following link: https://airtable.com/appFoFzjMcciPUgoK/api/docs#javascript/authentication
 
-Or by clicking on your user image in Airtable, and finding the "API Documentation" link:
+Or by clicking on HELP button in the upper right hand corner of Airtable, and finding the "API Documentation" link:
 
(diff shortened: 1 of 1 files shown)


M0nica:
I think the link that 404 is scoped to users. It probably would be helpful for you to update the docs to indicate that's a sample link vs a working link so that people can go head straight to walking through the steps to retrieve their own links/keys.

ifyoumakeit:
> I think the link that 404 is scoped to users. It probably would be helpful for you to update the docs to indicate that's a sample link vs a working link so that people can go head straight to walking through the steps to retrieve their own links/keys.

Sounds good! I can remove the link part and jump right to clicking the help link. I think the URL might be more confusing and the image gives a good idea of what the link should look like.

magnificode:
@ifyoumakeit - apologies for the delay in getting this merged! Appreciate your help in making the documentation nice and clear for new folks on the project :)

PR #274, opened 2020-06-24
Add skeleton loaders in businesses page

Mhmdabed11:
# Describe your PR
Add skeleton loaders in businesses page on initial page load and also while searching. Moreover, hide pagination while searching in businesses page.

The businesses page will be affected by this change.
- Before
https://www.loom.com/share/3ce5fb01db49464b800fbc52103cdc4f
- After
https://www.loom.com/share/51c0800875fc485883c3b7982e938106

## Steps to test

1. Go to `/businesses` page and refresh. You should see the skeleton loaders appearing before getting back the results.
2. Click on the search button to see the skeleton loaders appear while searching.

### DIFF
--- a/src/components/Feeds/BusinessFeed.js
+++ b/src/components/Feeds/BusinessFeed.js
@@ -20,6 +20,7 @@ function BusinessFeed({ businesses, onSearch, selectedFilters, loadingState }) {
     <Box
       maxW={theme.containers.main}
       paddingX={[null, theme.spacing.base, theme.spacing.lg]}
+      width="100%"
     >
       <BusinessFilter
         onSearch={filters => onSearch(filters)}
@@ -29,8 +30,16 @@ function BusinessFeed({ businesses, onSearch, selectedFilters, loadingState }) {
       {(searching || initialLoad) && (
         <Box mb={10}>
           <SimpleGrid columns={[null, 1, 2]} spacing={10}>
-            {[...Array.from(new Array(4))].map((val, index) => (
-              <Skeleton key={index} height="300px" />
+            {[...Array.from(new Array(20))].map((_, index) => (
+              <Skeleton key={index}>
+                <ResultCard
+                  category={{ id: 1, name: 'Entertainment' }}
+                  name="Saving Small Businesses In Chicago- Skyway Bowl"
+                  description="PlaceHolder Description"
+                  location="Placeholder Location"
+                  donationUrl="Placeholder Url"
+                />
+              </Skeleton>
             ))}
           </SimpleGrid>
         </Box>
--- a/src/components/Feeds/BusinessFeed.js
+++ b/src/components/Feeds/BusinessFeed.js
@@ -1,4 +1,4 @@
-import { Box, SimpleGrid, useTheme } from '@chakra-ui/core';
+import { Box, SimpleGrid, useTheme, Skeleton } from '@chakra-ui/core';
 import PropTypes from 'prop-types';
 import React from 'react';
 import NoResultsCard from '../Cards/NoResultsCard';
@@ -12,6 +12,8 @@ function BusinessFeed({ businesses, onSearch, selectedFilters, loadingState }) {
 
   const loaded = loadingState === LOADING_STATE.NONE;
   const initialLoad = loadingState === LOADING_STATE.INITIAL;
+  const searching = loadingState === LOADING_STATE.SEARCHING;
+
   const hasResults = businesses.length > 0;
 
   return (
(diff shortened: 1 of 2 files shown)


src/components/Feeds/BusinessFeed.js
```diff
@@ -24,7 +26,17 @@ function BusinessFeed({ businesses, onSearch, selectedFilters, loadingState }) {
         selectedFilters={selectedFilters}
         isSearching={loadingState === LOADING_STATE.SEARCHING}
       />
-      {!initialLoad && hasResults && (
+      {(searching || initialLoad) && (
+        <Box mb={10}>
+          <SimpleGrid columns={[null, 1, 2]} spacing={10}>
+            {[...Array.from(new Array(4))].map((val, index) => (
```
magnificode, line 32: @Mhmdabed11 is there any chance we can up the array value here to 20? Happy to discuss if you think that's not ideal. I figure since 20 is the number of results that we show per-page it might be a nice transition!

My comment below may negate this so read on 😄

src/components/Feeds/BusinessFeed.js
```diff
@@ -24,7 +26,17 @@ function BusinessFeed({ businesses, onSearch, selectedFilters, loadingState }) {
...
+              <Skeleton key={index} height="300px" />
```
magnificode, line 33: I believe that we can be a little bit more tricky with the `<Skeleton>` component here to better match the width and height of the incoming cards!

As per the documentation here: https://chakra-ui.com/skeleton I believe that we can actually wrap the `<ResultCard>` with a `<Skeleton>` component in order to have it take on the width and height of the incoming card. Did you play around with this by chance? It's entirely possible that this may not be the right solution, just wanted to point out the possibility!

src/components/Feeds/BusinessFeed.js
Mhmdabed11: Hello @magnificode , thank you for taking the time to check my pull request. Yes of course, there is no specific reason that I put 4, I just put it so it places a scroll on the page. But I think 20 makes more sense since, as you said, we are showing 20 results and it would be a nice transition.

src/components/Feeds/BusinessFeed.js
Mhmdabed11: I actually tried it. In our case, we have an image in the card which means that `<ResultCard>`, being a child of `<Skeleton>`,  should take a category for it to display an image skeleton, otherwise it will have no image and we will not get the desired height and width of the skeleton cards. What I can do here is the following:
1. Provide a category to `<ResultCard>` , for example `{ id: 1, name: 'Entertainment' }` which will make it take an image and thus showing a skeleton loader for the image.
2. The component which affects the width of the card and makes it take the width it currently does is  `<Heading>` in `<ResultCard>`. This is why I gave it a name, as shown below, which is long enough to make the 2 grid elements on the same row take the full width of the page which is `1220px`. 
3. I was observing a weird bug when providing `<ResultCard>`, as a child of `<Skeleton>`,  with a `websiteUrl` prop. The two anchors which show when given that prop were having a constant width which gave the page a horizontal scroll when resizing. Not providing the prop solved the problem. I am not sure exactly what caused it. 

Below is how the new loading skeleton might look like. 
```
 <SimpleGrid columns={[null, 1, 2]} spacing={10}>
            {[...Array.from(new Array(4))].map((_, index) => (
              <Skeleton key={index}>
                <ResultCard
                  category={{ id: 1, name: 'Entertainment' }}
                  name="Saving Small Businesses In Chicago- Skyway Bowl"
                  description="PlaceHolder Description"
                  location="Placeholder Location"
                  donationUrl="Placeholder Url"
                />
              </Skeleton>
            ))}
          </SimpleGrid>
```

magnificode: Yeah let's give that a shot if we can! Totally understandable if the amount of dynamic content fed to these cards is a bit too much to reliably get the inherited width and height. Worst case scenario we can set a `width` of 100%, a max width of whatever the value is for a card on desktop, and a height. 

Appreciate your thoroughness here!

Mhmdabed11: How about we give the parent `<Box>` component in  `<BusinessFeed>`  a `width` of `100%` ? It currently has a `maxWidth` of `1220px`. This way we do not give the skeleton cards any width, since their parent will take the entire width which is the same as its parent (in this case `1220px`). As for the height, I believe we cannot exactly get the exact value even if we set it to `100%` since the content is dynamically generated. I a more than happy to discuss further.

magnificode: Totally fine with that approach! If we can just get the skeleton cards to match the size as close as we can I'll be happy! Appreciate the discussion!!

magnificode:
This is excellent @Mhmdabed11 - thank you so much for working with me on this!!

Issue #280, opened 2020-06-26
[Feature] Add UTM to urls for Businesses


ashtine:
## User Story
As a business owner on RBB
I want to I want to see where traffic is coming from
So I can know how we're getting exposure

## Acceptance Criteria
- [ ] All external links to businesses include `utm_source=Rebuild+Black+Business` as query string
- [ ] Any link that already includes a query string properly has the UTM code appended.

## Is your feature request related to a problem? Please describe.
Add a tracking code to the end of any business URL so that the businesses themselves can see who is coming to their site from RBB

## Describe the solution you'd like
Dynamically append UTM tracking code to URLs with and without existing query params

JTSchwartz:
I'd be happy to work on this feature

PR #292, opened 2020-06-29
feat: added source tracking (utm) parameters to external business url…

JTSchwartz:
# Describe your PR

Dynamically adds utm source tracking queries to external business urls

Related to #280 

- /businesses ResultCards

RBB-PR-UTM

## Steps to test

1. Visit /businesses
2. Hover over or click on external links on ResultCards (Learn More, Donate)
3. Look at link preview or address bar 
    utm_source query will be set to "Rebuild+Black+Business" if one is not already given

### Additional notes

Works regardless of whether or not the url already has queries, and does not overwrite already established utm_source attributes. 

Links are verified to be external and use http/s protocol (i.e. no mailto links) before editing.

### DIFF
--- a/src/components/Card.js
+++ b/src/components/Card.js
@@ -16,7 +16,7 @@ function setUrlUtm(link) {
 
   if (
     // Ensure that the link is external and follows http protocol (as opposed to mailto)
-    url.hostname !== window.location.hostname &&
+    !url.hostname.includes('rebuildblackbusiness.com') &&
     url.protocol.includes('http')
   ) {
     const query = url.query;
--- a/src/components/Card.js
+++ b/src/components/Card.js
@@ -8,8 +8,30 @@ import {
   Text,
   Stack,
 } from '@chakra-ui/core';
-
+import URL from 'url-parse';
 import { Button, Image } from '.';
+
+function setUrlUtm(link) {
+  let url = new URL(link);
+
+  if (
+    // Ensure that the link is external and follows http protocol (as opposed to mailto)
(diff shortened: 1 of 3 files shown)


src/components/Card.js
```diff
@@ -8,9 +8,31 @@ import {
   Text,
   Stack,
 } from '@chakra-ui/core';
-
+import URL from 'url-parse';
 import { Button, Image } from '.';
 
+function setUrlUtm(link) {
```
magnificode, line 14: Can we move this function to our `utils` folder? We may end up wanting to use this in other places as well :)

src/components/Card.js
```diff
@@ -8,9 +8,31 @@ import {
   Text,
   Stack,
 } from '@chakra-ui/core';
-
+import URL from 'url-parse';
```
magnificode, line 11: Is there any way we can rely solely on the URL API? https://developer.mozilla.org/en-US/docs/Web/API/URL - the browser support is pretty much in line with what we need, and would allow us to eliminate the need for a dependency where we can!

I'm totally open to discussing this, if leveraging the URL API has some ramification I'm not thinking of that the `url-parse` package helps us solve, I'm all for it. Just figured I'd throw it out there!

src/components/Card.js
JTSchwartz: Can do

src/components/Card.js
JTSchwartz: Honestly, just never used that API and just straight for npm, glad you brought that up. I'll swap it over.

PR #298, opened 2020-07-01
add business registration promo to /businesses

Mhmdabed11:
# Describe your PR
Add business registration promo in `/businesses` page
Related to https://trello.com/c/kQ3Q4xcH

The `/businesses` page will be affected

image

## Steps to test

1. Go to `/businesses` page and check 4th card.

### DIFF
--- a/src/components/Feeds/BusinessFeed.js
+++ b/src/components/Feeds/BusinessFeed.js
@@ -36,6 +36,7 @@ function BusinessFeed({
         h="100%"
         top="0"
         left="0"
+        alt="looking-ip_whbkeh"
       />
       <CardContent
         color={theme.colors['rbb-black-200']}
--- a/src/components/Feeds/BusinessFeed.js
+++ b/src/components/Feeds/BusinessFeed.js
@@ -1,13 +1,11 @@
 import { Box, SimpleGrid, useTheme, Skeleton } from '@chakra-ui/core';
 import PropTypes from 'prop-types';
-import React, { useRef } from 'react';
+import React from 'react';
 import NoResultsCard from '../Cards/NoResultsCard';
 import BusinessFilter from '../Filters/BusinessFilter';
 import { LOADING_STATE } from '../../hooks/useSearch';
-import { CardWrapper, CardHeading, CardText, CardContent } from '../Card';
 import ResultCard from '../ResultCard';
-import Button from '../Button';
-import Image from '../Image';
+import RegistrationPromo from '../Promos/RegistrationPromo';
 
 function BusinessFeed({
   businesses,
@@ -16,7 +14,6 @@ function BusinessFeed({
   loadingState,
   onOpen,
 }) {
-  const focusRef = useRef();
   const theme = useTheme();
 
   const loaded = loadingState === LOADING_STATE.NONE;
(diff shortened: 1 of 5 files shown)


src/components/Feeds/BusinessFeed.js
```diff
@@ -48,10 +120,14 @@ function BusinessFeed({ businesses, onSearch, selectedFilters, loadingState }) {
       {!initialLoad && hasResults && !searching && (
         <>
           <SimpleGrid columns={[null, 1, 2]} spacing={10}>
-            {businesses.map(business => {
+            {businesses.map((business, index) => {
               const formattedLocation = `${business.city ? business.city : ''}${
                 business.city && business.state ? ', ' : ''
               }${business.state ? business.state : ''}`;
+
+              if (index === 3) {
```
magnificode, line 128: As it stands currently, this logic will overwrite the 4th business in the list.

Looking at the production site, the fourth business in the list is "Cloud 9 Bakers", this business (on the staging site here https://5efcde45a23084000711282c--rebuild-black-business.netlify.app/businesses) is overwritten by the `<RegistrationPromo>` component and does not show in the feed.

We need to make sure that the injection of this promo does not affect the order that businesses show up in the feed. I think that this may be something that the Backend team will need to help us with. So hold tight here for a bit @Mhmdabed11 while we figure that out.

@elchris @llexical - Is the API the thing that is determining the number of results that are returned to the frontend? If so, can we, with this PR, reduce that number to 19, to allow us to properly inject this promo component in such a way that the 4th business is also rendered and pagination is not negatively affected?

src/components/Feeds/BusinessFeed.js
```diff
@@ -16,6 +25,69 @@ function BusinessFeed({ businesses, onSearch, selectedFilters, loadingState }) {
 
   const hasResults = businesses.length > 0;
 
+  const RegistrationPromo = () => (
```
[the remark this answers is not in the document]
Mhmdabed11: @magnificode yeah of course, this makes perfect sense 👍 I will make sure it is in its own component file.

src/components/Feeds/BusinessFeed.js
Mhmdabed11: Oops, my bad. You are actually right, this will override the 4th result. The question is do we need to show the promo card on every page, and do we need to have an even number of cards? In both cases, if I return, in the conditional statement ‘if (index===3)’, both the promo code component and the ‘ResultCard’ component wrapped with a  react fragment, it will solve the problem of the overridden 4th result.

magnificode: @Mhmdabed11 no worries at all! Your line of thinking here is correct for sure. If we adjust this to output both the promo card, and the ResultCard we should be in good shape.

Additionally, we should reduce the `pageSize` parameter here https://github.com/Rebuild-Black-Business/RBB-Website/blob/ed5fb33ac7980db8afbc75cd6625153054f903e9/src/hooks/useSearch.js#L10 from `20` to `19` to make sure that the number of cards shown is always even.

magnificode:
@Mhmdabed11 this looks *excellent* thank you so so much for all the hard work, and for working with us while we figured out the best way to handle the logic :).

Mhmdabed11:
@magnificode Thank you for giving me the chance to work on this amazing project. This is actually my first open source project to contribute to, and it has been so much fun!

magnificode:
@Mhmdabed11 Ah!! That always warms my heart to hear! I can't thank you enough for being a part of the project, and I'm very glad that contributing here was a good experience for you 😄 !

PR #300, opened 2020-07-08
Ensure business links are absolute

JTSchwartz:
# Describe your PR

URLs are not verified to be in proper formatting, causing links enter without http/s protocol to attempt to route to a relative path within the rbb domain. This resolves the issue by ensuring an http:// protocol is present in the url

- /business

Before:
RBB-AbsoluteURL-Before

After: 
RBB-AbsoluteURL-After

## Steps to test

1. Navigate to /business
1. Hover over a link with a relative link (ex. Rebel Arts Movement Center)
1. Check URL preview or click on link

### DIFF
--- a/src/components/Card.js
+++ b/src/components/Card.js
@@ -8,7 +8,7 @@ import {
   Text,
   Stack,
 } from '@chakra-ui/core';
-import { setUrlUtm } from '../utils/urlUtils';
+import { setUrlUtm, verifyHttpUrl } from '../utils/urlUtils';
 import { Button, Image } from '.';
 
 const CardWrapper = forwardRef(({ children, ...props }, ref) => {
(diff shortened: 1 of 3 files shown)


magnificode:
Thanks for taking this on @JTSchwartz! The code looks solid but is failing deploy. Here's the error:

https://gyazo.com/b29960ee011bda6c837f7d0a465aec5f

It looks like `Placeholder URL` is being output somehow 🤔

JTSchwartz:
@magnificode thanks for pointing that out, could've sworn I saw green all the way down. 

I moved where the verification is called so that it isn't attempted by skeleton loaders (where "Placeholder URL" is coming from) and it looks like that's all it needed.

racedale:
@all-contributors please add JtSchwartz for code

Issue #281, opened 2020-06-26
[BUG] Get rid of the flash that happens when loading the website


ashtine:
**Describe the bug**
The flash that happens when loading the website

**To Reproduce**
1. Go to rebuildbackbusiness.com
2. Observe

**Expected behavior**
No flash and looks clean

**Actual Behavior**
A flash and kinda clunky

**Screenshots**
![Screen Capture on 2020-06-26 at 16-20-49](https://user-images.githubusercontent.com/25046304/85897853-2e38b800-b7c9-11ea-87fe-8817aa893300.gif)

ifyoumakeit:
I can take a crack at this, it looks like we can get a lot of mileage out of preloading the fonts and the hero image using meta tags in the <head />. Resizing the hero image using `srcSet` to fit will also speed up the site and remove the flash of it loading. 

Can use some dynamic URLS from cloudinary for srcSet https://cloudinary.com/documentation/responsive_images

ifyoumakeit:
Actually looking into it more, it looks like the hero image doesn't load without JavaScript.

magnificode:
@ifyoumakeit it might be useful to work with @domitriusclark on this one! Preloading fonts and other assets as best we can is a great idea as well. Feel free to move on what you can, and work with Dom on the image stuff in particular! I'll assign this task to you and him for now!

ifyoumakeit:
> @ifyoumakeit it might be useful to work with @domitriusclark on this one! Preloading fonts and other assets as best we can is a great idea as well. Feel free to move on what you can, and work with Dom on the image stuff in particular! I'll assign this task to you and him for now!

Sounds good!

magnificode:
@ifyoumakeit - @domitriusclark merged a few updates to the Cloudinary Image component today #299 . Not sure if that helps this issue, but figured I'd keep you updated :).

Anything we can do to help with this issue let us know!

JapneetSingh5:
Hi, I would love to contribute to your site. Can i comtribute to this issue or anywhere else you may need?

ifyoumakeit:
> @ifyoumakeit - @domitriusclark merged a few updates to the Cloudinary Image component today #299 . Not sure if that helps this issue, but figured I'd keep you updated :).
> 
> 
> 
> Anything we can do to help with this issue let us know!

Thanks for the update, I’ll see if this helps here.

magnificode:
[...]

PR #301, opened 2020-07-11
fix: white flash of before image load

ifyoumakeit:
This doesn't fix the overall problem but allows the background color
to show before the image comes in. This sets the background color
on ContentBlock. The z-index setting is also inverted to push
ContentBlockColorOverlay and ContentBlockContent on top of image
instead of pushing the image underneath.

# Describe your PR

Related to #281 

- Homepage, About

https://www.loom.com/share/a8cfec7efd854fa1b13462c6f10910fb
#281 has a video of the current loading 

## Steps to test

1. Load homepage and watch hero load in with black background instead of white.

### Additional notes

Doesn't fix completely just softens image entry, more work coming around load issues for #281

### DIFF
--- a/src/components/ContentBlock.js
+++ b/src/components/ContentBlock.js
@@ -24,6 +24,7 @@ const ContentBlockContent = props => {
       h="100%"
       w="100%"
       px={[4, 4, 8]}
+      zIndex={1}
       {...props}
     >
       {props.children}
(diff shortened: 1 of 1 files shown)


ifyoumakeit:
Not sure why the netlify preview doesn't get added, is that manual step for this repository?

Issue #313, opened 2020-07-26
[Feature] Adjust business listing grid


magnificode:
## User Story
As a user
I want to see 12 business listings per page
So I can more easily scan business listings

## Acceptance Criteria
- [ ] Business listing grid contains a max of 12 cards
- [ ] Each card sizes dynamically between 325px (min) and 450px (max)
- [ ] Each row should contain between 1 and 4 cards based on the defined min and max above.

## Describe the solution you'd like
The business listing grid is currently leveraging Chakra UI's SimpleGrid component. This should be adjusted to match the parameters listed above.

Mhmdabed11:
I can take this since I have worked before on the businesses page. Should we have 12 listings including the promo ?

magnificode:
@Mhmdabed11 thanks for picking this up! I think technically we'll want 11 listings + the promo, for a total of 12 cards in the grid. Let me know if you have any other questions!

Mhmdabed11:
@magnificode. In the below image, each card has a width of 325px (which is the min mentioned above). Below, I had to adjust the `max-width` of the page from `1220px` to `1518px` to account for the `325px` card width. I think that it became a bit crowded in this case.

 ![image](https://user-images.githubusercontent.com/39762577/88641933-8153a280-d0c8-11ea-922f-3732adec3972.png).

What I found looked good is the following

1. On large screen 
![image](https://user-images.githubusercontent.com/39762577/88642665-6fbeca80-d0c9-11ea-85b9-731e010f7395.png)

2. On medium screen
![image](https://user-images.githubusercontent.com/39762577/88642708-7fd6aa00-d0c9-11ea-9360-81e24ee188b8.png)

3. On mobile screen
![image](https://user-images.githubusercontent.com/39762577/88642793-9aa91e80-d0c9-11ea-890a-c8ba1dae8598.png)
Maybe this is not what you are looking for, I am happy to discuss further. Also please let me know if I am missing something here in the original requirements.

magnificode:
@Mhmdabed11 this is great, thank you so much for the thorough documentation. Let me pull @shapirone in here to comment on the top screen shot.

Looks like some of our listings with multiple buttons are looking a bit squished at this scale. Any thoughts on how we can address that?

shapirone:
Good catch @Mhmdabed11 - what's the min. width you're recommending for the card?

Mhmdabed11:
[...]

PR #319, opened 2020-08-03
adjust business listing grid

Mhmdabed11:
# Describe your PR

Adjust businesses listing grid to show 12 listings on the page, where each row has between 1 and 4 listings.

Related to #313 

`/businesses` page will be affected

1. 4 listings per row
image

2. 3 listings per row
image

3. 2 listings per row
image

4. 1 listing per row
image

## Steps to test

1. Go to `/businesses` page.
2. Resize window

### DIFF
--- a/src/gatsby-plugin-chakra-ui/theme.js
+++ b/src/gatsby-plugin-chakra-ui/theme.js
@@ -7,7 +7,7 @@ const customTheme = {
   ...theme,
   containers: {
     main: '1220px',
-    lg: '1516px',
+    lg: '1600px',
   },
   spacing: {
     xs: '0.75rem',
--- a/src/gatsby-plugin-chakra-ui/theme.js
+++ b/src/gatsby-plugin-chakra-ui/theme.js
@@ -7,6 +7,7 @@ const customTheme = {
   ...theme,
   containers: {
     main: '1220px',
+    lg: '1516px',
   },
   spacing: {
     xs: '0.75rem',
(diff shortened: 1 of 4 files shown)


src/gatsby-plugin-chakra-ui/theme.js
```diff
@@ -7,6 +7,7 @@ const customTheme = {
   ...theme,
   containers: {
     main: '1220px',
+    lg: '1516px',
```
magnificode, line 10: Feel free to up this to 1600px sir!

magnificode:
Hey @Mhmdabed11 - thanks again for this work. Couple small things and then I think we should be good to get this merged. Apologies for the delay in reviewing here, been a bit of a busy couple of weeks.

Screen Shot 2020-08-07 at 8 05 54 AM

Any chance we could add `hyphens: auto` to the headings here? That way if a longer title shows up in a more narrow card, the content will break onto a new line? You may need to add this as an inline `style`, as I'm not sure that Chakra has an attribute for this.

I added a comment about upping the container size to 1600 as well.

Appreciate the work!!

Mhmdabed11:
@magnificode Thank you for your comments and no worries at all! I had to add `wordBreak: break-all` instead of `hyphens: auto` since the `hyphens` property is not fully supported by all modern browsers.

magnificode:
@Mhmdabed11 good call! Any chance we can leverage both? That way modern browsers get the benefit of the hyphens, and older browsers will still see the word break?

Mhmdabed11:
@magnificode of course! Will do it in a few.

magnificode:
@Mhmdabed11 excellent work as always, thank you so much for the help!

Mhmdabed11:
@magnificode  You are most welcome!

racedale:
@all-contributors please add Mhmdabed11 for code

PR #320, opened 2020-08-07
Feature/individual business pages

### LINKED ISSUE #268: Add individual pages for each business listing
## Prerequisites
- [x] #267 :: The data structure needs to be solidified before we create these pages
- [x] Design for the profile page
- [x] [Backend endpoint to get a single building](https://trello.com/c/ZPKR6jit/37-add-endpoint-for-getting-a-business-by-id) 

## Designs
https://www.figma.com/file/Lo1ZcBOO9t4QRbX2I0PwBQ/RBB?node-id=2135%3A1
https://www.figma.com/file/Lo1ZcBOO9t4QRbX2I0PwBQ/RBB?node-id=2135%3A363

## The Problem

Currently users are only able to explore businesses found on the RBB site by the small amount of information we provide on the `ResultCard` components. We want to enable businesses to provide a wider range of information that we can leverage on a 'profile' page for each business.

## The Solution

We need to dynamically create a "Business Profile" template for each business in the database. Additionally, we need to conditional _all the things_ to account for our legacy data that will have much of this data missing, or incomplete. The page needs to be flexible enough to account for if the majority of this information is not available.

mbifulco:
# Describe your PR
This PR adds a URL for _each individual business_ sourced from Airtable.  As of right now, there's not much content on those pages (just some SEO stuff, and a title) - but the pages exist! Design implementation comes next.

**I'm opening this PR to get some feedback (please read below!)**

Fixes #268 

lots of new URLs!

image

## Steps to test

1. let this branch build (it may take a mighty long time to build on netlify the first time)
1. visit `/business/1946d007-3c76-5c4d-ac23-10e47bf759a5`

### TODO

- [x] It'd be great to use something other than an ID hash to generate URLs for businesses.  I created a function in `/utils/business.js` which should be modified to generate a slug for each business (and should have a strategy for duplicate slugs)
- [x] there's no content on individual business pages, that needs to be implemented from [figma](https://www.figma.com/file/Lo1ZcBOO9t4QRbX2I0PwBQ/RBB?node-id=2135%3A1)
- [x] we don't link to individual business pages anywhere


mbifulco:
Update - I should have the design implemented for this by the end of the weekend.  There'll be some loose ends that still need to be tied up, but we should be in a semi-releasable state soon!

mbifulco:
A few notes on this work:
- the current implementation is generating a route/page/URL for each business in the airtable _at build time,_ rather than pulling things from the API dynamically.  This decision was made to give individual businesses a leg up on SEO - these pages are inherently search engine crawlable and play nicely with robots.txt, etc.
- the current ID scheme (with big long GUIDs in the URL) should be changing to something much more human-readable, like `https://rebuildblackbusiness.com/mikes-candy-shop` where the ending slug is unique.  @elchris it would be super helpful if this were a unique field stored in the database somewhere.  Keeping IDs out of the URL is a best practice for SEO as well, per [moz](https://moz.com/learn/seo/url#:~:text=SEO%20best%20practices%20for%20URLs%20Keeping%20URLs%20as,is%20to%20use%20words%20that%20people%20can%20comprehend).
image

mbifulco:
I'm about to push a big ol' stack of changes to this PR.  It's taking shape nicely, but there's a bunch of data on the Profile page in Figma that we don't have yet.  That stuff is stubbed out here:

image

😅  I'm putting this down for the day... it's been a long sunday!

mbifulco:
# To do:
- [ ] update icons to use Font-Awesome Font.  Icon names are `store`, `globe`, `phone`, `share-alt`, `flag`, `share-square`
- [ ] call API after page load to populate additional metadata
- [ ] add social share links
- [ ] add share-this-page CTA
- [ ] update slugify to get rid of special characters like `'`

shapirone:
Looking good! A couple small design detail tweaks/feedback

- [ ] The image should be the same size as the business page (should be a little smaller and flush with the black background next to it)
- [ ] Donate text should be vertically centered
- [ ] We should likely have a character limit on the URL's display logic to make sure it's not taking up too much space (not as important for this one, but i'm dreading longer URLs)
- [ ] URL should have the underline that other links use
- [ ] Phone number and address should likely also be links (esp. for mobile devices)

mbifulco:
Allllright - I just pushed a mess of additional changes which address _some_ of the todo items above.  I think I'm still missing a bit of metadata from the API that may be needed to complete the page. This includes links to facebook/instagram/twitter/linkedin for businesses (yelp is present currently, from what I can see).  I'm also not sure what the `share profile` CTA button at the bottom of the page should do 🤔

mbifulco:
Ah hey I also remove yarn.lock to fix #330

mbifulco:
Shoot _one more thing_ - I added `react-query` as a dependency to the project to handle API fetching / caching / etc for individual business pages.  It works like a _dream_ and allows us to have a custom hook to make things work.  Because of that, you'll also see devtools for react-query when running the project locally (but not when it is built for production).  That looks like this:

image


and can be opened by clicking the little icon that sticks to the lower-left corner of the browser window:
image

src/templates/singleBusinessPage.js
```diff
@@ -0,0 +1,325 @@
...
+          >
+            Send us an email and let us know.
+          </Link>
+        </Text>
+      </Stack>
+    </Layout>
+  );
+};
+
+export const query = graphql`
```
juanri0s, line 303: this is more of a general comment than a suggestion, but doesn't it make more sense to turn these exported queries into hooks using `useStaticQuery` - similar to the linked doc below? 

https://www.gatsbyjs.com/docs/use-static-query/#composing-custom-usestaticquery-hooks

It can essentially help clean up some areas in our code and avoid the need to use the StaticQuery tag. This is obviously up for discussion so I'm happy to turn it into a GH issue instead.

src/templates/singleBusinessPage.js
```diff
@@ -0,0 +1,325 @@
...
+
+import { useImageForBusiness } from '../utils/business';
+import { useBusinessDetails } from '../hooks/useBusinessDetails';
+
+const SingleBusinessPage = ({ data }) => {
+  const theme = useTheme();
+
+  const business = data.airtableBusinesses.data;
+  const {
+    // name,
```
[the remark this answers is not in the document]
mbifulco: Hey @juanri0s - I left these commented out _mostly_ to let me know that these fields are available and yet unused as I'm getting this implementation done.  They should definitely get cleaned up before this gets merged!

src/templates/singleBusinessPage.js
mbifulco: This is a good idea too - I wonder if I can combine this with the hook I've written on this page to pull from the api...

mbifulco:
This is now hooked up to a search endpoint for airtable Ids, so the data for each page should be fully connected - check it out on the deploy preview: https://deploy-preview-320--rebuild-black-business.netlify.app/business/ashes-the-cigar-lounge!-rec4mmetqTsWYEAsE

juanri0s:
Desktop looks great, nice work!

Noticed a couple misalignments on the mobile preview, I also noticed that the category image takes a good amount of time to load, aren't they static assets? image

src/templates/singleBusinessPage.js
juanri0s: yeah, we can have a general metadata hook instead of writing different metadata queries but this is probably case-by-case. I'm okay leaving this implementation for another PR, but up to you.

src/templates/singleBusinessPage.js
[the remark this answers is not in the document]
juanri0s: yupp that makes sense, was just wondering!

In your most recent push, we get a warning for 

`'id' is assigned a value but never used` since you use `recordId` now

elchris:
i'm still seeing bad calls to the API on:

https://deploy-preview-320--rebuild-black-business.netlify.app/business/large-clothing-reciKTUpIWBFki1gw

/api/v1/businesses/airtable/null

* first issue: obviously, "null" where the air table ID should be
* 2nd issue: the URI is incorrect:
----- incorrect: /api/v1/businesses/airtable/null
----- correct:   /private/api/v1/businesses/airtable/reciKTUpIWBFki1gw

mbifulco:
Good catch @elchris  - making that change now.

elchris:
> Good catch @elchris - making that change now.

... it's still passing null to the endpoint instead of the airtable record id :(

mbifulco:
We're bumping into some problems getting the site to talk to the new API route - @elchris is on the case 👍

mbifulco:
Ok - we pushed an update to both this PR _and_ the API (both staging and production).  That _should_ do the trick.

src/templates/singleBusinessPage.js
```diff
@@ -65,6 +69,23 @@ const SingleBusinessPage = ({ data, pageContext }) => {
     zip,
   } = apiResponse || {};
 
+  const shareClicked = event => {
+    if (event && typeof event.preventDefault === 'function')
+      event.preventDefault();
+
+    debugger;
```
juanri0s, line 76: won't this pause the program?

mbifulco: Yikes! Yes that needs to go.  Good catch!

Issue #289, opened 2020-06-27
[Feature] Create "Fundraisers" page


magnificode:
## Prerequisites
- [x] Design

Figma Designs: https://www.figma.com/file/Lo1ZcBOO9t4QRbX2I0PwBQ/RBB?node-id=2008%3A105

## User Story
As a user
I want to view a page that only shows fundraisers
so I can quickly view ways to support black owned business

## Acceptance Criteria
- [ ] A new top level Fundraisers page should be added to the top level navigation
- [ ] Businesses that have a "Donate" entry in the database should be pulled out of the Business Listing feed, and shown on the newly created "Fundraisers" page.
- [ ] Businesses that have a "Donate" entry in the database must also show in the main business listing page.

## Describe the solution you'd like
We want to highlight businesses in need without them getting lost in the entire business directory. Separating Businesses with a "Donate" link from the main business listing page should help with this.

Mhmdabed11:
I can start on this issue if nothing is on hold!

magnificode:
@Mhmdabed11 that would be excellent, thank you so much!

Mhmdabed11:
@magnificode 
![image](https://user-images.githubusercontent.com/39762577/90382956-2807df00-e088-11ea-9274-3682de371279.png)
I have a list of questions regarding this issue: 
- Where can I find the public Id of the image above
- What should happen when the button in the card above is pressed?
- Currently I am fetching businesses using a Graphql query on Airtable businesses collection . I cannot find the location/state attribute which should show in the fundraiser card.

elchris:
Please don’t query AirTable directly, use our REST API for this:

https://api.rebuildblackbusiness.com/

Use the “Find Businesses” GET endpoint:

https://api.rbb-api.com/api/v1/businesses?in-need=true

And be sure to pass the x-api-key header whose value should come from a local .env file , which in your dev environment can be your personal api key from the portal and in production will be automatically replaced with our production key. You can see the code example from the main “See Businesses” site.

Mhmdabed11:
Okay @elchris  thank you for your input !!

magnificode:
Merged in #321

PR #321, opened 2020-08-18
Feature/create fundraisers page

Mhmdabed11:
# Describe your PR
Create the fundraisers page
Related to #289 

A new fundraisers page will be created

## Steps to test

1. Navigate to /fundraisers

### DIFF
--- a/src/components/Cards/FundraiserCard.js
+++ b/src/components/Cards/FundraiserCard.js
@@ -1,0 +1,96 @@
+import React, { forwardRef, useRef } from 'react';
+import { Box, Icon, Text, useDisclosure, useTheme } from '@chakra-ui/core';
+import PropTypes from 'prop-types';
+import {
+  CardButton,
+  CardButtonGroup,
+  CardContent,
+  CardHeading,
+  CardText,
+  CardWrapper,
+} from '../Card';
+import Link from '../Link';
+import ContactModal from '../ContactModal';
+
+/**
+ * @component
+ * @example
+ * return (
+ *   <FundraiserCard
+ *     name="Jane Example"
+ *     donationLink="https://somedonation.link"
+ *   />
+ * )
+ */
+
+const FundraiserCard = forwardRef(({ name, donationLink, ...props }, ref) => {
+  const { onOpen, isOpen, onClose } = useDisclosure();
+  const { updateRef } = useRef();
+  const theme = useTheme();
+
+  return (
+    <>
+      <CardWrapper ref={ref} {...props}>
+        <CardContent
+          bg={theme.colors['rbb-white']}
+          color={theme.colors['rbb-black-100']}
+          display="flex"
+          flexDirection="column"
+        >
+          <CardHeading
+            fontFamily={theme.fonts['heading-slab']}
+            textTransform="uppercase"
+            fontSize={theme.fontSizes.xl}
+            lineHeight="1"
+            overflowWrap="break-word"
+            wordWrap="break-word"
+            wordBreak="break-word"
+            hyphens="auto"
+          >
+            {name}
+          </CardHeading>
+          {/* {location && (
+            <CardText
+              as="p"
+              fontFamily={theme.fonts.heading}
+              fontSize={theme.fontSizes.lg}
+            >
+              {location}
+            </CardText>
+          )} */}
+
+          <Box marginTop="auto" paddingTop={theme.spacing.base}>
+            <CardButtonGroup pb={3}>
+              <CardButton as="a" href={donationLink}>
+                Donate
+              </CardButton>
+            </CardButtonGroup>
+            <Text as="small" fontSize="sm" mt={3} isInline>
+              <Icon
+                name="flag"
+                color={theme.colors['rbb-gray']}
+                mr={theme.spacing.xs}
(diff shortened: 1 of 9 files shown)


Mhmdabed11:
@magnificode  I have some questions regarding this PR so it can be 100% done from my side:
1. Is there a way to filter the businesses which have a donationLink entry using the REST API. I cannot fins such one in the api docs.
2. image
For the above image, what should happen when we press on the button? And also I need the public ID of the background image if possible.
3.For this page, I think it makes sense to have more than 12 entries in the grid since the size of each gridItem is relatively small. This means that we need to edit the useSearch hook to accept an optional pageSize argument. I can do this if we agree to do it.

src/components/Cards/FundraiserCard.js
```diff
@@ -0,0 +1,96 @@
+import React, { forwardRef, useRef } from 'react';
+import { Box, Icon, Text, useDisclosure, useTheme } from '@chakra-ui/core';
+import PropTypes from 'prop-types';
+import {
+  CardButton,
+  CardButtonGroup,
+  CardContent,
+  CardHeading,
+  CardText,
```
magnificode, line 9: We can remove this import if CardText is not used :)

src/components/Cards/FundraiserCard.js
```diff
@@ -0,0 +1,96 @@
...
+            fontSize={theme.fontSizes.xl}
+            lineHeight="1"
+            overflowWrap="break-word"
+            wordWrap="break-word"
+            wordBreak="break-word"
+            hyphens="auto"
+          >
+            {name}
+          </CardHeading>
+          {/* {location && (
```
magnificode, line 52: Remove commented code here

magnificode:
> @magnificode I have some questions regarding this PR so it can be 100% done from my side:
> 
> 1. Is there a way to filter the businesses which have a donationLink entry using the REST API. I cannot fins such one in the api docs.
> 2. image
>    For the above image, what should happen when we press on the button? And also I need the public ID of the background image if possible.
>    3.For this page, I think it makes sense to have more than 12 entries in the grid since the size of each gridItem is relatively small. This means that we need to edit the useSearch hook to accept an optional pageSize argument. I can do this if we agree to do it.

Excellent questions @Mhmdabed11 

1. We should be able to, maybe @elchris can shed some light there?
2. Public ID for that image: `coffee-shop_rbgwyx` - @shapirone, could you speak to the functionality for that CTA?
3. It looks like the designs have 14 items output on this page. I think adding that optional argument would be excellent!

Last thing, would you mind removing the "Home" link from the nav? The additional fundraisers page makes the nav a bit long.

Screen Shot 2020-08-20 at 11 31 32 AM

Based on the design, the order should be as follows:

Screen Shot 2020-08-20 at 11 40 15 AM


Thank you so much for the hard work on this @Mhmdabed11 - this is looking great!

Mhmdabed11:
@magnificode  Thank you for answering my questions and you are most welcome !

elchris:
oh my bad to get businesses raising funds, pass:

in-need=true

It's been documented on the API Doc:

https://developer.rbb-api.com/doc

click on the 3rd endpoint:

GET /api/v1/businesses

@Mhmdabed11 @magnificode

Mhmdabed11:
@elchris thank you for your reply. I actually did read this in the docs, but I ended up with some businesses with `donationLink:null` while setting `in-need=true`.

shapirone:
@Mhmdabed11 sorry I wasn't clear but since we support fundraisers that aren't directly attached to businesses, we have a different (and much simpler) UX for adding a fundraisers vs businesses. Designs can be found in [our Figma.](https://www.figma.com/file/Lo1ZcBOO9t4QRbX2I0PwBQ/RBB?node-id=1983%3A8264)

@magnificode do we have a separate issue for creating the Fundraiser form or should that be handled as part of this ticket?

magnificode:
@shapirone @Mhmdabed11 lets focus on displaying the existing businesses with donation links in this PR.

We'll create another task for submitting a fundraiser not connected to a business.

magnificode:
@Mhmdabed11 this looks awesome, and great touch pre-checking the "Has Fundraiser" checkbox when submitting from this page. I think that this will work great for now, we can iterate on the data structure on the backend further down the line if we want to split this out into it's own table.

The final, tiny thing, if you wouldn't mind @Mhmdabed11 - could we add a couple of transforms to this image? As it stands, the image itself makes the text on top a bit tough to read.

Screen Shot 2020-08-24 at 10 39 23 AM

In order to match the design, I added a couple of transformations to the image on cloudinary:

```
background: "#000000"
effect: "saturation:-78"
opacity: "47"
```

The Cloudinary Image takes a `transformations` object as props, which you should be able to pass these values to, in order to see the result, which should look similar to this: https://res.cloudinary.com/rebuild-black-business/image/upload/b_rgb:000000,e_saturation:-78,o_47/v1597945031/assets/coffee-shop_rbgwyx.jpg

Let me know if you're able to get that to work!

Thank you again for all the hard work! Once that image is handled we can get this merged in :)

magnificode:
Awesome work @Mhmdabed11 merging this in now!

PR #377, opened 2020-11-11
Implement client side routing and data fetching for individual businesses 

### LINKED ISSUE #374: [BUG] / [FEATURE] Convert "Business Profiles" to simple Single-Page App, Give-up SEO for Now, and Change URL Scheme to Reference API ID instead of AirTable ID
**NOTE**: This issue would replace #372 and #371 
**NOTE**: All API URLs which I pasted below will not work without an **?apiKey=** parameter appended to them. You can obtain your own API key instantly by visiting and registering at: https://api.rebuildblackbusiness.com and from there you can actually [read the API doc](https://developer.rbb-api.com/doc) from which you can explore the various API endpoints.

- The pre-generation of pages was done to promote SEO. But we never did quite get it to work properly as described in #372 -- So as of right now, we literally have no vested interest in SEO.

- The pre-generation of pages:
  - adds overhead in development and deployment because it takes a whole hour to statically generate every single page for over 13,000 businesses. In development, we have to operate against a different instance of the API which has a smaller data set. Doing this however results in not being able to fully test the experience, most especially "searches", because there literally is very little to nothing to search for.
  - is very likely to be the root of #371 , which could potentially be fixed by adding some sort of hook whenever a new business gets added to AirTable

- the pre-generated pages are happening by loading data from AirTable and that’s going to continue to hurt us because it forces us to reference the AirTable ID in the urls instead of referencing a business’s actual [API](https://api.rebuildblackbusiness.com) ID from the API database which is what the developer API does .

- In the end we need to remove our dependence on AirTable. It needs to remain at arm’s length.

What I propose:

- Convert the business profile to a full SPA loading ALL of [its data from the API](https://api.rebuildblackbusiness.com), not AirTable. The single page app should only interact with the API. It should not touch AirTable. As such, no more pre generation of pages will be needed.

- change the business profile url to be anchored against the [integer API ID](https://api.rbb-api.com/api/v1/businesses/100) instead of the AirTable ID.  Keep the "textual part" of the slugs the same: Here's what I mean by example:
  - **Before**: https://www.rebuildblackbusiness.com/business/cassy's-support-service-llc-recNXBoD6kZPC5kZf
    - Notice the "recNXBoD6kZPC5kZf" part, that's the identifier for the AirTable record. We need to get away from that.
    - The API is able to retrieve records by AirTable ID right now, but it's a "private" feature which I want to disable
    - This the API call for it: https://api.rbb-api.com/private/api/v1/businesses/airtable/recNXBoD6kZPC5kZf
  - **After (This is what we want)**: https://www.rebuildblackbusiness.com/business/cassy's-support-service-llc-13499
    - Notice the "13499": that's the API ID: https://api.rbb-api.com/api/v1/businesses/13499

- Once the business profile url scheme has been updated, and the profile page works correctly for any ID, be sure to update [links in the search results pages](https://www.rebuildblackbusiness.com/businesses/all?page=3#results).

Benefits:

- get rid of issues #371 and #372 
- reduce our dependency on AirTable for customer-facing experiences, and instead stay architecturally coupled to the API
- reduce development and deployment overhead with a simpler architecture
- should allow us to move more easily to NextJS later, whereby all of the React logic built to render the Business Profile page in an SPA paradigm, gets invoked and run on-demand using NextJS's ability to dynamically render a React page on-demand. See #375 for a description of NextJS-based approach.
- actually make the Owners Portal possible to finish. Because I need to make it possible for an owner to create a new business which does not exist in airtable. I will be saving the business in the application with an API integer ID, and it will not exist in AirTable so it will not have an AirTable id, and as such there is currently no way to link to it using an API ID. And I thought about enhancing the sync from a one-way sync (AirTable toward API) to a two-way sync (add API to AirTable) and when you play various use cases scenarios, it becomes a hot mess. Once an an owner claims a business or adds their own listing, it becomes severed from AirTable, and that’s how it should be.

Mhmdabed11:
This PR implements client side routing and data fetching for individual businesses. So when a new business is added, users can directly visit its profile page without having to trigger a gatsby build.

Related to #374 
Fixes #374 

No design changes.  The url of the profile business page will now hold the id of the business in the database rather than the airtable id.

## Steps to test

1. Register a new business
2. Visit the profile page of the business immediately and it should be working.

### Additional notes
I set the `refetchOnWindowFocus` attribute of `React Query` to false since the business details in the profile page are being refetched on window focus and it was a bit misleading especially that we do not have a placeholder while the page is loading. This shows alot when the business does not exist, which in this case RQ will refetch several times before throwing an error. I am more than happy to discuss this further in case it has some use-case I am not aware of.

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -15,39 +15,47 @@ exports.onCreatePage = async ({ page, actions }) => {
   }
 };
 
-exports.createPages = async ({ graphql, actions }) => {
-  const result = await graphql(`
-    query {
-      businesses: allAirtableBusinesses {
-        nodes {
-          data {
-            Business_Name
-            Name
-          }
-          recordId
-        }
-      }
-    }
-  `);
-
-  result.data.businesses.nodes.forEach(business => {
-    const slug = getSlugForBusiness({
-      businessName: business.data.Business_Name,
-      name: business.data.Name,
-      airtableId: business.recordId,
-    });
-
-    if (slug) {
-      actions.createPage({
-        path: `/business/${slug}`,
-        component: require.resolve('./src/templates/singleBusinessPage'),
-        context: {
-          businessId: business.recordId,
-        },
-      });
-    }
+exports.createPages = ({ actions }) => {
+  actions.createPage({
+    path: `/business`,
+    matchPath: `/business/:id`,
+    component: require.resolve('./src/templates/singleBusinessPage'),
   });
 };
+
+// exports.createPages = async ({ graphql, actions }) => {
+//   const result = await graphql(`
+//     query {
+//       businesses: allAirtableBusinesses {
+//         nodes {
+//           data {
+//             Business_Name
+//             Name
+//           }
+//           recordId
+//         }
+//       }
+//     }
+//   `);
+
+//   result.data.businesses.nodes.forEach(business => {
+//     const slug = getSlugForBusiness({
+//       businessName: business.data.Business_Name,
+//       name: business.data.Name,
+//       airtableId: business.recordId,
+//     });
+
+//     if (slug) {
+//       actions.createPage({
+//         path: `/business/${slug}`,
+//         component: require.resolve('./src/templates/singleBusinessPage'),
+//         context: {
+//           businessId: business.recordId,
+//         },
+//       });
+//     }
+//   });
+// };
 
 // switch off type inference for SitePage.context
 // more info here - https://www.gatsbyjs.com/docs/scaling-issues/#switch-off-type-inference-for-sitepagecontext
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -1,5 +1,3 @@
-const { getSlugForBusiness } = require('./src/utils/business');
-
 // Implement the Gatsby API “onCreatePage”. This is
 // called after every page is created.
 exports.onCreatePage = async ({ page, actions }) => {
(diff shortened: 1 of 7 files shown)


gatsby-node.js
```diff
@@ -15,40 +15,48 @@ exports.onCreatePage = async ({ page, actions }) => {
...
-    }
+exports.createPages = ({ actions }) => {
+  actions.createPage({
+    path: `/business`,
+    matchPath: `/business/:id`,
+    component: require.resolve('./src/templates/singleBusinessPage'),
   });
 };
 
+// exports.createPages = async ({ graphql, actions }) => {
```
magnificode, line 26: Feel free to fully remove this commented out code if it's no longer necessary!

magnificode (approved):
@Mhmdabed11 excellent work as always! One small note about removing some commented code, but otherwise looks awesome. Thank you!!

magnificode:
@Mhmdabed11 it does appear that we're getting a build error when testing this deployment:

Screen Shot 2020-11-11 at 8 37 39 AM

Mhmdabed11:
> @Mhmdabed11 it does appear that we're getting a build error when testing this deployment:
> 
> Screen Shot 2020-11-11 at 8 37 39 AM

Ohh I guess that is because `splitSlug` does not exist on the server. I will look into that

PR #381, opened 2020-11-15
Adds basic form validation

### LINKED ISSUE #379: [BUG] "Register A Business" Form: Make Protocol Mandatory on ALL URLs
**Describe the bug**
When filling out information to Register a Business, a user is still allowed to enter URLs without a protocol. We must enforce that client side.

**To Reproduce**
1. Go here https://www.rebuildblackbusiness.com/businesses
2. Click on 'Register a Business'
3. Scroll down to 'Website'
4. Notice that while it suggests that you enter a protocol, it does not **enforce it**
5. Check the box that says "Business has an ongoing donation campaign"
6. Notice the same issue with "Donation Link" field

**Expected behavior**
No form should be allowed to be submitted without all URLs having a protocol in front of them

**Actual Behavior**
www.foo.com is allowed and 99% of the time what people put in there.

**Epilogue and Broader Context**
- Note that this should be a low-hanging quick fix to the issue at hand, and very-much well worth doing.
- Eventually this work could be further improved by implementing #323 
  - We could keep the input form exactly the same, but instead of submitting to AirTable, it would submit to the RBB API with the following benefits:
    - The API gives back helpful and detailed validation messages when certain conditions are not met
    - We can stop leaking our AirTable API Credentials in the client side code and http requests. This is **not** a good place to be.

brianjhanson:
This started out as a request to help users remember to put `http://` and `https://` in their website addresses. This also adds a rough system for validating forms in general. 

When a user submits the form, we run all of the values through a `validateSignUpForm` function which will return errors in an object with each input's key as a property. Something like:
```js
{
    website: 'Please enter a valid URL',
    email: 'Please enter a valid email', 
    name: 'Name is required',
}
```

We then set this object to an `errors` state and use that to check when fields have errors we need to show. 

Fixes #379 

* Register a business form

**Error when no protocol is present**
Error when no protocol is present

**Updated error display throughout the form**
Updated error display throughout the form

**Gif of error experience**
Gif of error experience

## Steps to test

1. Go to the businesses page and click "Register a Business"
2. Fill out the website field without a protocol and submit

### Additional notes
* In order to show errors correctly I had to wrap each form item in a `FormControl` element which updates the structure of the form a bit
* I updated the form to actually be a form component which should provide better accessibility and reduce a tiny bit of the complexity around it
* I think we should consider using [Formik](https://formik.org/docs/overview) and [Yup](https://github.com/jquense/yup) for this kind of thing down the road. Happy to open an issue for that if there's a desire for it. 
* I made some minor updates to the visuals of the form, increasing the size of the modal and adjusting the spacing around each form "row"

### DIFF
--- a/src/components/Forms/BusinessSignUpForm.js
+++ b/src/components/Forms/BusinessSignUpForm.js
@@ -99,7 +99,7 @@ const BusinessSignUpForm = ({ isFundraiser = false }) => {
       return;
     }
 
-    // submitBusiness(infoToSubmit);
+    submitBusiness(infoToSubmit);
     setSubmitted(true);
   };
 
--- a/src/components/Forms/BusinessSignUpForm.js
+++ b/src/components/Forms/BusinessSignUpForm.js
@@ -153,7 +153,7 @@ const BusinessSignUpForm = ({ isFundraiser = false }) => {
   }
 
   return (
-    <form onSubmit={handleSubmit} method="POST" noValidate>
+    <form onSubmit={handleSubmit} method="POST">
       {/* renders when form is submitted with validation errors */}
       {validationMessage && (
         <Box bg="red.50" p={4} color="red.500">
--- a/src/components/Forms/BusinessSignUpForm.js
+++ b/src/components/Forms/BusinessSignUpForm.js
@@ -15,11 +15,14 @@ import {
   Icon,
   RadioGroup,
   Radio,
+  FormErrorMessage,
+  Box,
 } from '@chakra-ui/core';
 
 import PrimaryButton from '../Buttons/PrimaryButton';
 import { submitBusiness } from '../../services/AirtableServices';
 import { stateObj } from '../../utils/stateObj';
+import { validateSignUpForm } from '../../utils/validateSignUpForm';
 
 const businessCategories = [
   'Entertainment',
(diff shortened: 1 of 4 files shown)


src/components/Forms/BusinessSignUpForm.js
```diff
@@ -143,105 +148,110 @@ const BusinessSignUpForm = ({ isFundraiser = false }) => {
...
-      maxWidth="1000px"
-      margin="0 auto 3rem"
-      padding="0 24px"
-      onKeyPress={event => {
-        if (event.key === 'Enter') {
-          handleSubmit(event);
-        }
-      }}
-    >
+    <form onSubmit={handleSubmit} method="POST" noValidate>
```
brianjhanson: Wrapping the form in an actual `form` removes the need to listen for a specific key press. It also should provide a slightly more accessible experience.

magnificode (approved):
@brianjhanson this looks _excellent_! Thank you very much for the work here.

I'm all for leveraging a framework like Formik to make form management and validation a bit less painful. I think that's something we can explore as we re-assess our tech stack in the near future. Building those new forms with Formik from the ground up will definitely make things easier.

Please feel free to open up an issue and we can discuss that. I'll happily tag that issue as something we need to address in the next iteration of the site!

brianjhanson:
@magnificode added a ticket for Formik + Yup [over here](https://github.com/Rebuild-Black-Business/RBB-Website/issues/384#issue-749305436). My first story here and it's not really a user focused story so please don't hesitate to let me know if I should add more detail.

PR #382, opened 2020-11-18
Search from home page

### LINKED ISSUE #353: [Feature] Make homepage landing searchable
## User Story
As a user
I want to search for a black owned business through the home page
So I can quickly find the information I'm looking for without having to navigate to another page.

## Acceptance Criteria
- [ ] Search bar, and associated header indicating that a user can search for black owned businesses
- [ ] On submit, the user is taken to the business page, with their query shown and results ready/pending
- [ ] More or less, think of this like the google home page 

## Is your feature request related to a problem? Please describe.
It is not always immediately clear to users that they can search for black owned businesses on our site. We want to prioritize this functionality and make it apparent as soon as a user lands on the home page.

## Describe the solution you'd like
We want the user to hit the homepage and instantly have the ability to search for black-owned businesses w/o having to click around. This will lead to a better user experience by giving the user what they want in a quicker fashion. An easy approach to this could be to have a styled search box on our homepage that when a user enters their search, their values are transferred over to the "Business" search page via Query String Params, and immediately executed on page load with the data.

brentmclark:
# Describe your PR

Closes #353 

#### Before

image

#### After
image

## Steps to test

1. Open home page
2. Observe the fact that you can now search
3. Perform a search
4. Observe that you are now on the business search page with search results

### Additional notes

I added a lightweight variant setup to the `BusinesFilter` component which can now be used to display either of the two options.  The existing filter was intended to display _on dark_ backgrounds; this new one is intended for light backgrounds.

Also, this _appears_ the be the first instance of programmatic navigation in the app, unless I've overlooked something.  Please let me know if I've chosen the incorrect approach.

#### Regarding the design

I chose what you see here because it's obvious enough to be immediately useful (affordance) while also not detracting _too_ much from the primary intent of the landing page.

### DIFF
--- a/src/components/Feeds/BusinessFeed.js
+++ b/src/components/Feeds/BusinessFeed.js
@@ -37,11 +37,13 @@ function BusinessFeed({
       paddingX={[null, theme.spacing.base, theme.spacing.lg]}
       width="100%"
     >
-      <BusinessFilter
-        onSearch={filters => onSearch(filters)}
-        selectedFilters={selectedFilters}
-        isSearching={loadingState === LOADING_STATE.SEARCHING}
-      />
+      <Box margin="0 auto 3rem">
+        <BusinessFilter
+          onSearch={filters => onSearch(filters)}
+          selectedFilters={selectedFilters}
+          isSearching={loadingState === LOADING_STATE.SEARCHING}
+        />
+      </Box>
       {(searching || initialLoad) && (
         <Box mb={10}>
           <SimpleGrid columns={[null, 1, 2, 3, 4]} spacing={10}>
--- a/src/components/Feeds/BusinessFeed.js
+++ b/src/components/Feeds/BusinessFeed.js
@@ -37,7 +37,7 @@ function BusinessFeed({
       paddingX={[null, theme.spacing.base, theme.spacing.lg]}
       width="100%"
     >
-      <Box margin="0 auto 3rem">
+      <Box margin="0 0 3rem 0">
         <BusinessFilter
           onSearch={filters => onSearch(filters)}
           selectedFilters={selectedFilters}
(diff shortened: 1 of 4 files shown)


src/components/Feeds/BusinessFeed.js
```diff
@@ -37,11 +37,13 @@ function BusinessFeed({
       paddingX={[null, theme.spacing.base, theme.spacing.lg]}
       width="100%"
     >
-      <BusinessFilter
-        onSearch={filters => onSearch(filters)}
-        selectedFilters={selectedFilters}
-        isSearching={loadingState === LOADING_STATE.SEARCHING}
-      />
+      <Box margin="0 auto 3rem">
```
brentmclark, line 40: This was me adopting a "No outside white space" policy on this component. The internal application of the `3rem` bottom margin was making component reuse difficult.

magnificode (commented):
@brentmclark this looks awesome man, thanks so much for taking this on!

There is one tiny tweak that I'd like to see on the actual business page:

https://gyazo.com/0ed6d2f4d3494f647f2ca513797a70df

The max width on the component is preventing it from being full width, and centered here. If we could adjust that so it's centered then this puppy will be good to push to prod.

Really appreciate the hard work here!

brentmclark:
> @brentmclark this looks awesome man, thanks so much for taking this on!
> 
> There is one tiny tweak that I'd like to see on the actual business page:
> 
> https://gyazo.com/0ed6d2f4d3494f647f2ca513797a70df
> 
> The max width on the component is preventing it from being full width, and centered here. If we could adjust that so it's centered then this puppy will be good to push to prod.
> 
> Really appreciate the hard work here!

Thanks for catching that @magnificode.  I got a little too greedy when I move the margin around and removed the `margin: auto` that was keeping it centered.  Want to have another look to ensure everything looks correct on your end?

